Issue #143, opened 2026-05-22
Publisher field - More than one publisher


HansUwe52:
At present, only one publisher can be entered in the “Publisher” field. If you try to add a further publisher, the first one is simply deleted and replaced by the second.
However, there are many publications that have more than one publisher. 
Therefore, the field should be changed so that multiple publishers can be entered  – in the same way as is the case with the author field.

fabiodalez-dev:
Implementato in PR #148 (branch `feat/multi-publisher-143`).

Commit collegati:
- a1f83102 — multi-publisher core (schema `libri_editori` + migrazione 0.7.15, form Choices.js multi-select, save, display, 10 test E2E)
- de0deff9 — export interoperabili (OAI-PMH MARC21/UNIMARC/MODS/Dublin Core/MAG + BIBFRAME) estesi a tutti gli editori

`libri.editore_id` resta come editore primario per retro-compatibilità; l'elenco completo vive nella junction `libri_editori`.

PR #156, opened 2026-05-30
fix: multi-publisher merge/filters, PHP 8.2 preflight floor, release verify

fabiodalez-dev:
## Summary

Addresses four code-review findings (2× P1, 2× P2) plus pre-existing semgrep noise on touched files.

### P1 — PHP 8.1 install/update could break at Composer bootstrap
`composer.json` requires `php: ^8.2` and the generated `vendor/composer/platform_check.php` fatally exits below 8.2, but the updater and installer still accepted 8.1 — an 8.1 host could pass preflight, install the release, then die at the Composer bootstrap. Floor raised to `8.2.0` in `Updater::checkRequirements()`, `Validator::validateSystemRequirements()` and installer step 1 (label updated too).

### P1 — Publisher merge dropped multi-publisher associations
`libri_editori` (issue #143) has a FK `ON DELETE CASCADE` to `editori`, so merging away a duplicate silently wiped its junction rows. `PublisherRepository::mergePublishers()` now repoints those rows onto the primary (`INSERT IGNORE ... SELECT`) **before** the `DELETE FROM editori` fires the cascade. Guarded for installs predating the migration (table absent → falls back to `editore_id` only).

### P2 — Publisher views/filters ignored secondary publishers
CSV export, the public publisher archive, and the publisher-scoped repositories filtered only on `libri.editore_id`. Switched to `(l.editore_id = ? OR EXISTS(SELECT 1 FROM libri_editori ...))` so a book also appears under its secondary publishers, without regressing primary matches:
- `LibriController` export filter
- `FrontendController` publisher archive (count + list)
- `BookRepository::getByPublisherId()`
- `PublisherRepository::getBooksByPublisherId()`

### P2 — Release smoke test did not verify bundled plugin `frbr-lrm`
Added `frbr-lrm` to `create-release.sh` `BUNDLED_PLUGINS` so it matches `BundledPlugins::LIST` and the ZIP cannot ship without it.

### semgrep cleanup (pre-existing, on touched files)
- `openssl_decrypt` failure now handled explicitly (fail closed) in `Updater`.
- `unlink-use` audit findings triaged with `// nosemgrep` + justification on provably-internal paths (updater-controlled `.maintenance`/`.lock`/zip/temp paths; log-rotation `glob()` confined to the log dir; cover-delete already has a `realpath()` containment guard).

## Testing
- `php -l` clean on all 7 changed PHP files
- PHPStan level 5: **No errors**
- `bash -n scripts/create-release.sh`: ok
- semgrep rescan of `Updater.php` + `LibriController.php`: **0 findings**
- DB functional test (transaction + rollback, real data intact):
  - P2a: EXISTS filter finds a book via its **secondary** publisher (old `editore_id`-only filter did not — regression reproduced then resolved)
  - P1b: merge repoints the association onto the primary; duplicate's junction rows correctly removed by cascade
- `BundledPlugins::LIST` vs `create-release.sh`: aligned (empty diff)

No schema change → no new migration required.

## Summary by CodeRabbit

* **New Features**
  * Supporto multi-editore esteso: filtri, conteggi, archivi, esportazioni e import mantengono libri collegati anche come co-editori; sincronizzazione automatica della relazione primaria.

* **Chores**
  * Requisito minimo PHP aggiornato a 8.2; aggiornamenti ai manifest plugin e script di creazione release (inclusione frbr-lrm).
  * Nuova utilità per l’introspezione schema per compatibilità pre/post-migrazione.

* **Tests**
  * Nuove suite E2E per multi-editore, multi-autore e plugin Archives.

* **Migration**
  * Aggiunta migration idempotente per la tabella di giunzione multi-editore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,30 @@ Pinakes is a self-hosted, full-featured ILS for schools, municipalities, and pri
 
 ---
 
+## What's New in v0.7.16
+
+### Multi-publisher, hardened end to end (#143)
+
+Books can have more than one publisher (the `libri_editori` junction, introduced in 0.7.15). This release closes every gap in that model: publisher **filters, counts, exports, the public publisher archive, the catalog facet, search, the admin API and bulk operations** now all match a book whether the publisher is its primary one (`libri.editore_id`) or a secondary one in the junction. Merging two publishers re-points the junction onto the survivor **before** the cascade, so no association is lost; the publisher-delete guard counts secondary links too; and CSV / LibraryThing import and bulk-enrichment now keep the junction in sync so interop exporters (OAI-PMH, BIBFRAME) never lose a publisher.
+
+Every new junction query is **guarded for pre-migration installs** — on a database that predates the junction table the queries gracefully fall back to the primary publisher instead of erroring.
+
+### PHP 8.2 is now the floor
+
+The installer and the in-app updater now require **PHP 8.2+**, matching `composer.json` (`^8.2`) and the generated `platform_check.php`. Previously an 8.1 host could pass preflight and then die at the Composer bootstrap.
+
+### Other fixes
+
+- **Multi-character book-case codes** (#153): the legacy single-letter UNIQUE constraint on `scaffali.lettera` is dropped, so codes like `L1`, `L2` no longer collide.
+- **Edit form**: the "Import from ISBN" field is pre-filled with the book's ISBN/EAN when editing.
+- A reconciliation migration heals any `libri_editori` drift left by imports written before the sync landed.
+
+### Testing
+
+The comprehensive E2E suite grew to **132 tests**, adding a 20-test **Archives** phase (ISAD(G) CRUD, hierarchy, SQL seeding, authority records, and the JSON/XML APIs — RiC-O JSON-LD, IIIF, OAI-PMH, SRU, MARCXML/Dublin Core/EAD3/METS) and a 9-test **multi-publisher / multi-author** phase. Validated with a fresh-install + real-upgrade regression.
+
+---
+
 ## What's New in v0.7.14
 
 ### Installer fix: wizard no longer wedges at step 6 (Configurazione Email)
--- a/app/Controllers/CsvImportController.php
+++ b/app/Controllers/CsvImportController.php
@@ -1332,6 +1332,36 @@ class CsvImportController
         $stmt->close();
 
         $this->syncImportedSeries($db, $bookId, $collana, $numeroSerie);
+        $this->syncPrimaryPublisherJunction($db, $bookId, $editorId);
+    }
+
+    /**
+     * Keep libri_editori in sync with a book's primary publisher (issue #143).
+     *
+     * CSV import writes libri.editore_id directly (not via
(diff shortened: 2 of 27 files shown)


fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
## Deep review — multi-publisher coverage (rev. `ad0e345f`)

Multi-lens pass over the branch diff (correctness / structural blast-radius / project-conventions / comments / UX / security). Range `main..HEAD`. Below: every candidate and its disposition. All actionable findings are fixed in this branch.

### Fixed in this branch

| Sev | Area | Finding | Fix |
|-----|------|---------|-----|
| High | Catalog filter | `FrontendController::buildWhereConditions()` matched only the **primary** publisher (`e.nome = ?`) while the sidebar facet count already counted secondaries → count/results mismatch (secondary-only book in the badge but not in the list) | `(e.nome = ? OR EXISTS(libri_editori … e2.nome = ?))` |
| High | Interop | CSV import, LibraryThing import and bulk-enrichment write `libri.editore_id` **without** populating `libri_editori`, so junction-only consumers (OAI-PMH, BIBFRAME) export no publisher for imported/enriched books | additive `INSERT IGNORE … ordine 0` sync at every `editore_id` write (6 sites, 3 files) |
| Low | Merge | After repointing junction rows, the primary's `ordine=0` marker could be lost when the book already listed the primary as a secondary | re-assert `ordine=0` for the surviving primary post-repoint |
| Trivial | Docs | `countBooks()` comment said the bulk-delete relies on it; the bulk-delete has its own inline guard | comment corrected |

### Reviewed — no change needed

- **SQL injection / binds** — all new query fragments use parameterized placeholders; the doubled `IN ($placeholders)` binds the int-sanitized id list twice correctly; every added `?` binds the id the right number of times. Clean.
- **`nosemgrep` on `unlink()`** — each annotated path is app-internal (constants under `rootPath`, log-dir `glob()`, realpath-contained cover path); no genuine path-traversal suppressed.
- **`openssl_decrypt` fail-closed** — empty/null on failure flows into existing empty-key guards (token skipped, SRU 401); no auth-bypass or leak.
- **Publisher archive display label** — a secondary-only book on a publisher's archive shows the book's *primary* publisher name; this is valid book metadata (the book's main publisher), not false data — left as-is by design.
- **`Installer.php` open-library seed `requires_php` 8.1 → 7.4** — intentional: aligns the seed with the plugin's own manifest (`requires_php` is the plugin minimum, not the app floor, which is enforced separately at 8.2).
- **CLAUDE.md compliance** — soft-delete `AND deleted_at IS NULL` preserved on every broadened `libri` query; no hardcoded routes; migration version ≤ release. Clean.

### Verification
- `php -l` + PHPStan level 5 — clean on all touched files
- semgrep — 0 findings on touched files
- transaction-rollback DB tests asserting every new query shape (filter / count / guard / junction-sync / ordine) returns/affects the right rows, with primary-only counter-proofs
- full E2E suite (`full-test.spec.js`) — **103 passed / 0 failed** (covers CSV·TSV import + ISBN/scraping enrichment)

PR #167, opened 2026-06-09
Complete backup system: DB + uploaded files, with restore (#162)

### LINKED ISSUE #162: Backup incomplete
Backup system only export SQL from database, but do not export uploads directory (at least covers...)

fabiodalez-dev:
## Complete backup system (Closes #162)

The backup previously dumped only the database SQL. It now produces a single
ZIP that bundles the **database dump + the uploaded files**, and can **restore**
both — the rollback path the system is meant for (e.g. after a dirty XLS import).

### What's included
- **Backup as a ZIP** (`storage/backups/backup_<ts>.zip`): `database.sql` +
  `manifest.json` (version, scope, table/file counts, SHA-256) +
  `files/public/uploads/…` and `files/storage/uploads/plugins/…`. **`.env` is
  never included.** Legacy `update_*/database.sql` backups stay listable.
- **Scope choice** when creating a manual backup: full (DB + files) or DB-only.
- **Restore** (admin-only, destructive): from a stored backup or an uploaded ZIP.
  A pre-restore **safety backup** is always created first; the DB import runs on
  a dedicated connection; file extraction is **zip-slip guarded**.
- **Pre-update backup** stays DB-only by default, with an admin setting to make
  it full.
- **Streaming download** (no longer loads the whole file into memory).

### Security
- Restore is **admin-only** (not staff), CSRF-guarded, with strong confirmations.
- Zip-slip guard on extract; `.env` never backed up or written; uploaded-archive
  size cap + manifest validation; path-traversal guards on every backup name.

### UI / i18n
- Reuses the existing Tailwind chrome, SweetAlert + CSRF patterns. New "Contenuto"
  column and Ripristina action, an admin-only "Ripristina da file" upload, and the
  pre-update toggle. Every new string is translated in all four locales (it/en/de/fr).

### Verification
- `tests/backup-restore.spec.js` (E2E): full backup contents (incl. no `.env`),
  ZIP download, **restore round-trip** (deleted row + deleted cover file both come
  back, safety backup created), staff session rejected (403), delete. **5/5 green.**
- PHPStan L5 clean; pre-update backup delegate verified.

## Summary by CodeRabbit

* **New Features**
  * Backup configurabile (solo DB o completo) con checkbox persistente per includere file; UI admin estesa (upload ZIP, selezione scope).
  * Ripristino admin-only da ZIP salvati o caricati (operazione distruttiva, CSRF, conferma); backup di sicurezza pre-restore; restore parziale segnalato.
  * Download in streaming con header corretti; upload ZIP limitato a 2 GB; operazioni legacy non ripristinabili.

* **Tests**
  * Suite e2e Playwright per create/download/restore/upload-restore/permessi/delete.

* **Localization**
  * Nuove stringhe per backup/ripristino (de,en,fr,it).


fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
### Code review

**Branch:** `feature/162-complete-backup` → `main`
**PR:** #167 (open)
**Review ID:** `rev_20260609T200240Z715548`
**Sub-agent tokens:** 3,119,582 across 48 invocations

Found 35 findings across all lanes:
- Deep lane (correctness/security): 11 resolved

---

### ✓ Auto-fixable (11)

| # | Score | Impact | File | Issue | Status |
|---|-------|--------|------|-------|--------|
| F011 | 70 | correctness | `app/Support/BackupManager.php:330-349` | In restoreZip() the irreversible DB import (importDatabase, step 4) and the post-import promoteStagedFiles() (step 5) share one catch: a throw from promoteStagedFiles AFTER the DB has already been replaced returns success=false with a generic error, so the user is told the restore failed even though the database was already overwritten — they may retry or restore the safety backup over a partially-applied state. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`80ae16e`) |
| F013 | 35 | correctness | `app/Support/BackupManager.php:500-504` | dumpDatabaseTo() serializes every column value via real_escape_string()+single-quote regardless of type; binary/BLOB column bytes are emitted as a quoted text literal (no _binary/hex literal), so any binary data stored in the DB is corrupted on backup/restore round-trip. Same flaw existed in the removed Updater::backupDatabase but this PR re-introduces it as the now-canonical dump path used for restore. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`889ce06`) |
| F014 | 45 | correctness | `app/Support/BackupManager.php:857-909` | extractFilesSafely() streams each ZIP entry to disk with no cap on cumulative decompressed size; an admin-uploaded restore ZIP is only size-capped on its compressed form, so a decompression-bomb archive can exhaust disk during restore and leave a half-promoted filesystem. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`889ce06`) |
| F015 | 40 | correctness | `app/Support/BackupManager.php:617-648` | importViaCli() treats any non-zero exit as a clean retry signal and falls through to the PHP importer, but `mysql < dump` aborts on the first failing statement having already executed the preceding DROP/CREATE/INSERTs; recovery relies on DROP TABLE IF EXISTS idempotency, but a CLI failure mid-trigger can leave the DB in a state the fallback does not fully reset. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`889ce06`) |
| F019 | 70 | security | `app/Controllers/UpdateController.php:202-231` | deleteBackup() performs no tipo_utente === 'admin' re-check, so staff users (admitted by AdminAuthMiddleware) can delete any stored backup; the sibling restoreBackup() docstring explicitly notes this gap and adds the re-check, but deleteBackup() — an equally destructive operation — was left without it. | ✓ fixed and verified (`80ae16e`) |
| F033 | 62 | security | `app/Controllers/UpdateController.php:122-150` | createBackup endpoint has no admin role re-check — any authenticated staff user can trigger a potentially large full-backup (DB + all uploads), creating a resource-exhaustion risk, since the backup ZIP is stored under storage/backups and later downloadable. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`80ae16e`) |
| F034 | 45 | security | `app/Support/BackupManager.php:621-629` | addslashes() is used to escape credential values written into a MySQL option file (.cnf), but addslashes() is not the correct escaping for MySQL option-file format — a backslash in DB_PASS or DB_USER (e.g. 'pa\ss') becomes a double backslash which MySQL interprets differently, causing authentication failure or value injection if the password contains special sequences. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`889ce06`) |
| F036 | 45 | security | `app/Support/BackupManager.php:637` | exec() output ($out) contains the merged stderr of the mysql CLI invocation (2>&1), which may include the full error message and is silently discarded — no logging occurs, so a failed CLI import is completely invisible in the audit trail. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`889ce06`) |
| F038 | 40 | security | `app/Support/BackupManager.php:243-260` | The 2 GB upload size cap is enforced after the file has already been fully received by PHP (moveTo has completed) — an attacker can upload a 2 GB+ file and exhaust disk space before the check triggers. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`889ce06`) |
| F039 | 35 | security | `app/Support/BackupManager.php:302-303` | The manifest validation only checks that the 'app' key equals 'Pinakes' — an attacker who can supply an uploaded ZIP (admin-only, but still) could craft a manifest with app=Pinakes and a crafted database.sql containing arbitrary DROP/CREATE/INSERT statements, and the whole dump will be imported; the only integrity check is the SHA-256 which is also attacker-controlled in the manifest. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`889ce06`) |
| F041 | 45 | security | `app/Support/BackupManager.php:679-716` | importViaPhp loads the entire SQL dump into memory via file_get_contents — for a large database this could exhaust PHP memory and kill the process mid-restore, leaving the database in a destroyed half-imported state with no rollback possible (DDL is auto-committed). <sub>(human-confirmed)</sub> | ✓ fixed and verified (`889ce06`) |

<details><summary>Details and fix proposals</summary>

#### F011 — In restoreZip() the irreversible DB import (importDatabase, step 4) and the post-import promoteStagedFiles() (step 5) share one catch: a throw from promoteStagedFiles AFTER the DB has already been replaced returns success=false with a generic error, so the user is told the restore failed even though the database was already overwritten — they may retry or restore the safety backup over a partially-applied state.

**File:** `app/Support/BackupManager.php:330-349`
**Score:** 70 (moderate)
**Human-confirmed:** @auto-rec/fabiodalezbackup@gmail.com at 2026-06-10T06:01:48Z — auto-accepted via :fix Phase 7.5 preflight
_Promoted from disposition=`manual` / actionability=`manual` / score_phase4=70_
**Fix direction:** In `app/Support/BackupManager.php`, add a `$dbImported = false;` flag before the try block and set it to `true` immediately after line 324 (`$this->importDatabase(...)`). In the catch at line 335, branch on `$dbImported`: when `true`, return `['success'=>false,'partial'=>true,'restored_phase'=>'database','safety_backup'=>$safetyName,'error'=>$e->getMessage()]` instead of the plain failure shape. In `app/Controllers/UpdateController.php` lines 286-293 and 334-341, forward `partial` and `restored_phase` from `$result` into the JSON response. In `app/Views/admin/updates.php` `runRestore()` (~line 1161), add a `data.partial` branch before the generic error path that fires a `warning`-icon SweetAlert instructing the user not to retry and to restore only the safety backup if needed.

**Evidence:**
- app/Support/BackupManager.php:324 — importDatabase($sqlTmp,...) (step 4) runs DROP TABLE + recreate with no surrounding transaction across DDL; after this line the live DB is already the restored one (comment line 322-323 confirms 'After this point the DB is the restored one')
- app/Support/BackupManager.php:486 — dump emits 'DROP TABLE IF EXISTS' and importViaCli/importViaPhp replay it, so the import is irreversible in-place (no rollback of the live schema)
- app/Support/BackupManager.php:330 — promoteStagedFiles($stagingDir) (step 5) runs AFTER the DB is replaced and can throw RuntimeException on mkdir/rename/copy failure (lines 389,401,409)
- app/Support/BackupManager.php:335-344 — a single catch wraps BOTH step 4 and step 5; a throw from promoteStagedFiles returns ['success'=>false,'safety_backup'=>$safetyName,'error'=>$e->getMessage()] — identical shape to a pre-import failure, with no marker that the DB was already overwritten
- app/Controllers/UpdateController.php:293 / :341 — both restoreBackup and uploadRestore relay $result['error'] with HTTP 500 on success=false; no partial-restore signal is forwarded
- app/Views/admin/updates.php:1161-1166 — runRestore treats any non-success as a flat error ('Errore durante il ripristino'); user is told the restore failed with no hint that the database was already replaced and only file promotion failed
- app/Support/BackupManager.php:243 & :226 — both restoreFromUploadedZip and restoreFromBackup funnel through restoreZip, so the issue affects both restore entry points

**Approach:** Split the single catch so that failures after the irreversible DB import are reported distinctly from pre-import failures. Track a phase marker inside restoreZip (e.g. $dbImported = false set true immediately after importDatabase succeeds). In the catch, if $dbImported is true, return a distinguishable result: ['success'=>false,'partial'=>true,'restored_phase'=>'database','safety_backup'=>$safetyName,'error'=>...] naming exactly what was applied (DB replaced, file promotion incomplete). Surface 'partial' through both UpdateController methods, and in runRestore() render a warning-level message ('Il database è stato ripristinato ma il ripristino dei file è fallito — NON riprovare il ripristino; ripristina il backup di sicurezza solo se necessario') instead of the flat error. Optionally make promoteStagedFiles best-effort/collecting errors so a single file failure doesn't abort the whole promotion after the DB is already committed.

**Files to modify:**
- `app/Support/BackupManager.php` — Add a $dbImported flag set true right after importDatabase() (line 324); in the catch (335-344) branch on it and return a partial-restore shape (add 'partial' + 'restored_phase' keys) when the throw occurred after the DB was replaced (why: The DB import is irreversible; the result must tell the caller the DB was already applied so the failure is not indistinguishable from a no-op pre-import failure)
- `app/Controllers/UpdateController.php` — Forward the new 'partial'/'restored_phase' fields in both restoreBackup (286-293) and uploadRestore (334-341) JSON responses (why: The controller currently drops everything except error/safety_backup, so the UI can never learn the DB was applied)
- `app/Views/admin/updates.php` — In runRestore() (1138-1170) detect data.partial and show a warning icon + explicit guidance not to retry and how to recover via the safety backup, instead of the generic error (why: Prevents the user from re-running the destructive import or blindly restoring the safety backup over a half-applied state)

**Verification:**
- Unit/integration: force promoteStagedFiles to throw (e.g. make a live upload root unwritable / inject a failing rename) AFTER importDatabase succeeds, and assert the returned array has success=false plus partial=true / restored_phase='database'
- Assert the pre-import failure path (e.g. corrupt checksum at importDatabase line 585) still returns success=false WITHOUT the partial flag — the two must remain distinguishable
- E2E (tests/backup-restore.spec.js): simulate a file-promotion failure and assert the UI shows the partial-restore warning, not the flat 'Errore durante il ripristino'

**Edge cases to preserve:**
- Pre-import failures (safety backup creation, invalid manifest, missing database.sql, checksum mismatch, staging extraction) must still report a plain non-partial failure since the live DB is untouched
- Successful full restore must still return success=true with safety_backup and no partial flag
- $safetyName must continue to be returned in all failure cases so the user can recover
- Temp/staging cleanup (unlink sqlTmp / removeDir stagingDir) must still run on the partial-failure path

**Latest fix attempt (fixrun_20260610T055559Zae5b41):** fixed and verified

#### F013 — dumpDatabaseTo() serializes every column value via real_escape_string()+single-quote regardless of type; binary/BLOB column bytes are emitted as a quoted text literal (no _binary/hex literal), so any binary data stored in the DB is corrupted on backup/restore round-trip. Same flaw existed in the removed Updater::backupDatabase but this PR re-introduces it as the now-canonical dump path used for restore.

**File:** `app/Support/BackupManager.php:500-504`
**Score:** 35
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-06-10T10:03:12Z — walkthrough: F013 — picked option C (Document the no-BLOB invariant in code comment)
_Promoted from disposition=`?` / actionability=`auto_fixable` / score_phase4=None_
**Fix direction:** In app/Support/BackupManager.php dumpDatabaseTo(), above the array_map value-serialization block (~line 511, locate by content), add the comment: `// Schema invariant: no BLOB/BINARY columns — real_escape_string quoting is safe for all text/numeric/datetime types.` Also add a `@throws`-style note to the method docblock stating binary columns are not supported by this dump path. Do NOT add runtime checks, do NOT change the serialization logic — comment-only change.

**Latest fix attempt (fixrun_20260610T114937Z0f7a11):** fixed and verified

#### F014 — extractFilesSafely() streams each ZIP entry to disk with no cap on cumulative decompressed size; an admin-uploaded restore ZIP is only size-capped on its compressed form, so a decompression-bomb archive can exhaust disk during restore and leave a half-promoted filesystem.

**File:** `app/Support/BackupManager.php:857-909`
**Score:** 45
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-06-10T07:46:39Z — walkthrough: F014 — picked option A (Add MAX_RESTORE_DECOMPRESSED_BYTES constant and track cumulative bytes written)
_Promoted from disposition=`uncertain` / actionability=`manual` / score_phase4=45_
**Fix direction:** Add `private const MAX_RESTORE_DECOMPRESSED_BYTES = 4 * 1024 * 1024 * 1024;` after MAX_UPLOAD_BYTES in app/Support/BackupManager.php. In extractFilesSafely(), initialise `$totalDecompressed = 0;` before the loop, replace `stream_copy_to_stream($stream, $out);` with `$written = stream_copy_to_stream($stream, $out); $totalDecompressed += $written;` and immediately after add `if ($totalDecompressed > self::MAX_RESTORE_DECOMPRESSED_BYTES) { fclose($stream); fclose($out); throw new \RuntimeException(__('Archivio di backup non valido: dimensione decompressa eccessiva')); }`. New user-facing string must be added to all 4 locale JSONs per project rule. Do NOT add per-entry statIndex pre-flight, do NOT pre-scan the archive — cumulative counter only.

**Latest fix attempt (fixrun_20260610T114937Z0f7a11):** fixed and verified

#### F015 — importViaCli() treats any non-zero exit as a clean retry signal and falls through to the PHP importer, but `mysql < dump` aborts on the first failing statement having already executed the preceding DROP/CREATE/INSERTs; recovery relies on DROP TABLE IF EXISTS idempotency, but a CLI failure mid-trigger can leave the DB in a state the fallback does not fully reset.

**File:** `app/Support/BackupManager.php:617-648`
**Score:** 40
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-06-10T10:03:13Z — walkthrough: F015 — picked option A (Strengthen the docblock to document trigger idempotency explicitly)
_Promoted from disposition=`?` / actionability=`auto_fixable` / score_phase4=None_
**Fix direction:** In app/Support/BackupManager.php importViaCli()/importDatabase docblock (~line 619-626, locate by content 'DROP TABLE IF EXISTS makes a retry idempotent'), replace that sentence with: 'the PHP fallback then re-imports cleanly because the dump is fully idempotent: every table is preceded by DROP TABLE IF EXISTS (which also cascades to owned triggers), and the triggers section additionally emits DROP TRIGGER IF EXISTS before each CREATE TRIGGER — so a partial CLI abort leaves no object the fallback cannot cleanly overwrite.' Documentation-only; do NOT change control flow; do NOT duplicate the SecureLogger warning being added by F036.

**Latest fix attempt (fixrun_20260610T114937Z0f7a11):** fixed and verified

#### F019 — deleteBackup() performs no tipo_utente === 'admin' re-check, so staff users (admitted by AdminAuthMiddleware) can delete any stored backup; the sibling restoreBackup() docstring explicitly notes this gap and adds the re-check, but deleteBackup() — an equally destructive operation — was left without it.

**File:** `app/Controllers/UpdateController.php:202-231`
**Score:** 70 (moderate)

**Evidence:**
- app/Controllers/UpdateController.php:202-231 — deleteBackup() validates only CSRF + non-empty backup name, then calls BackupManager::deleteBackup($backupName). No tipo_utente === 'admin' re-check.
- app/Controllers/UpdateController.php:264-272 — restoreBackup() docstring: 'Destructive → admin-only (AdminAuthMiddleware also lets staff through, so re-check here)' and gates on (($_SESSION['user']['tipo_utente'] ?? '') !== 'admin') → 403. Same guard in uploadRestore() (301-303) and saveBackupSettings() (349-351).
- app/Middleware/AdminAuthMiddleware.php:46-47 — middleware admits BOTH 'admin' and 'staff' ($tipo_utente !== 'admin' && $tipo_utente !== 'staff'). So staff reach every backup endpoint unless the controller re-checks.
- app/Routes/web.php:3060-3064 — /admin/updates/backup/delete is guarded only by CsrfMiddleware + AdminAuthMiddleware (no per-route admin-only filter), so staff pass through to deleteBackup().
- app/Routes/web.php:3042-3046 — /admin/updates/backup (createBackup) likewise CsrfMiddleware + AdminAuthMiddleware only; createBackup() in the controller (diff lines ~132-150) has no tipo_utente re-check either.
- git show main:app/Controllers/UpdateController.php — on main the only tipo_utente re-check in the controller is checkAvailable (read-only gate, line 168); deleteBackup existed on main WITHOUT a re-check, so staff could already delete backups pre-PR. The PR newly added the admin-only pattern for restore/upload/settings but did not retrofit delete/create.
- app/Support/BackupManager.php:178-198 — deleteBackup() resolves the name via resolveBackup() (realpath under storage/backups, no traversal) and unlinks; i.e. the operation is genuinely destructive (removes a recovery point) once reached.

**Approach:** Apply the same admin-only re-check the PR added to restoreBackup/uploadRestore/saveBackupSettings to the remaining destructive/sensitive backup endpoints. Concretely add, as the first statement in each handler (before CSRF/body parsing), the guard: `if (($_SESSION['user']['tipo_utente'] ?? '') !== 'admin') { return $this->jsonResponse($response, ['error' => __('Operazione riservata agli amministratori')], 403); }`. Cover deleteBackup (destructive, primary), createBackup (resource/disk), and downloadBackup (exfiltration of full DB+uploads). getBackups (listing only) is lower risk — gate it too for consistency, or document the deliberate exception. Prefer the controller-level re-check over widening AdminAuthMiddleware so staff keep access to non-backup admin features.

**Files to modify:**
- `app/Controllers/UpdateController.php` — Add the admin-only tipo_utente !== 'admin' → 403 guard at the top of deleteBackup() (~line 204), createBackup(), and downloadBackup() (~line 238); optionally getBackups() (~line 192). (why: deleteBackup destroys recovery points; createBackup consumes disk; downloadBackup exfiltrates a full DB+files archive. All are currently reachable by staff because AdminAuthMiddleware admits staff and these handlers omit the re-check the PR applied to restore/upload/settings.)
- `app/Routes/web.php` — Optionally add clarifying comments on routes 3042 (backup create), 3060 (backup/delete), 3066 (backup/download) mirroring the existing 3078-3079 comment, noting the controller enforces admin-only. (why: Keeps route-level documentation consistent with the restore/upload routes and signals the deliberate admin-only intent.)

**Verification:**
- As a 'staff' user (session tipo_utente='staff'), POST /admin/updates/backup/delete with a valid CSRF token and an existing backup name → expect 403 with 'Operazione riservata agli amministratori' and the backup still present on disk.
- Repeat for POST /admin/updates/backup (create) and GET /admin/updates/backup/download → expect 403 for staff.
- As an 'admin' user, the same requests must still succeed (200) — confirm no regression for the privileged role.
- Confirm restoreBackup/uploadRestore/saveBackupSettings continue to return 403 for staff and 200 for admin (unchanged).

**Edge cases to preserve:**
- CSRF validation must still run and return 403 on invalid token for admins (order: admin-check first like the sibling methods, or after — match the restore methods which check admin first).
- Empty/missing backup name → keep 400 for admins.
- Do NOT widen AdminAuthMiddleware to drop staff globally — staff legitimately use other admin views; the fix must be per-endpoint.

**Latest fix attempt (fixrun_20260610T055559Zae5b41):** fixed and verified

#### F033 — createBackup endpoint has no admin role re-check — any authenticated staff user can trigger a potentially large full-backup (DB + all uploads), creating a resource-exhaustion risk, since the backup ZIP is stored under storage/backups and later downloadable.

**File:** `app/Controllers/UpdateController.php:122-150`
**Score:** 62 (moderate)
**Human-confirmed:** @auto-rec/fabiodalezbackup@gmail.com at 2026-06-10T06:01:48Z — auto-accepted via :fix Phase 7.5 preflight
_Promoted from disposition=`manual` / actionability=`manual` / score_phase4=62_
**Fix direction:** Two-part fix: (1) In `app/Routes/web.php` lines 3042-3046, append `->add(new \App\Middleware\RateLimitMiddleware(5, 60))` to the `POST /admin/updates/backup` route, matching the rate-limit pattern on other update routes at line 3022. (2) In `app/Controllers/UpdateController.php`, add the same admin re-check already present at lines 301-303 to `createBackup()` (line ~124), `downloadBackup()` (~line 238), and `deleteBackup()` (~line 204) so all three endpoints are gated to admin-only, consistent with `restoreBackup()`/`uploadRestore()`.

**Evidence:**
- app/Controllers/UpdateController.php:122-150 — createBackup() has comment 'Admin-only access check removed' and no in-controller role re-check; only validates CSRF.
- app/Controllers/UpdateController.php:135 — default scope is 'full' when 'scope' param absent: (($data['scope'] ?? 'full') === 'db') ? 'db' : 'full'.
- app/Controllers/UpdateController.php:136 — calls new BackupManager(...)->createBackup($scope); full scope zips DB + uploads.
- app/Middleware/AdminAuthMiddleware.php:46 — allows BOTH 'admin' and 'staff' (tipo_utente !== 'admin' && !== 'staff' → deny), so staff pass through.
- app/Routes/web.php:3042-3046 — POST /admin/updates/backup is guarded only by CsrfMiddleware + AdminAuthMiddleware; NO RateLimitMiddleware (unlike /admin/updates:3022, /check:3028, /available:3034, /history:3052, /token:3110).
- app/Controllers/UpdateController.php:268-272 & 299-303 — restoreBackup() and uploadRestore() DO add an explicit admin-only re-check ('!== admin' → 403) with comment 'AdminAuthMiddleware also lets staff through, so re-check here', proving deliberate awareness; createBackup/downloadBackup/deleteBackup were intentionally left at staff level.
- app/Support/BackupManager.php:15-32,55-88 — full backup includes files/public/uploads/... and files/storage/uploads/plugins/... zipped to storage/backups/backup_<ts>.zip.
- git show main:app/Support/Updater.php:1824 — pre-PR endpoint called Updater::createBackup() which backed up DB only (database.sql, no uploads); BackupManager did not exist on main → PR genuinely widens per-call disk/CPU cost.
- app/Controllers/UpdateController.php:236-262 — downloadBackup() also lacks an admin re-check, so staff can download the produced full DB+uploads archive.

**Approach:** Two-part: (1) Add RateLimitMiddleware to POST /admin/updates/backup to cap how often a full backup can be triggered (resource-exhaustion mitigation), matching the rate-limiting already applied to other update routes. (2) Decide the role policy explicitly: if backup creation/download are meant to be admin-only like restore, add the same in-controller admin re-check used at lines 270/301 to createBackup(), downloadBackup(), and deleteBackup(); if staff backup is intentional, keep create open but still rate-limit, and strongly consider gating downloadBackup() to admin since it exfiltrates a full DB dump + all uploads. The maintainer should confirm intent (per project memory project_cover_handling_and_uploads_decisions / backup roadmap), but the rate-limit fix is unambiguous and safe.

**Files to modify:**
- `app/Routes/web.php` — Append ->add(new \App\Middleware\RateLimitMiddleware(N, 60)) to the POST /admin/updates/backup route (lines 3042-3046); pick a small N (e.g. 5/min) since full backups are heavy. (why: Caps staff-triggered full-backup frequency, removing the resource-exhaustion/DoS vector the candidate flags.)
- `app/Controllers/UpdateController.php` — If policy is admin-only: add the existing admin re-check pattern (if (($_SESSION['user']['tipo_utente'] ?? '') !== 'admin') return 403) to createBackup() (line ~124), and at minimum to downloadBackup() (line ~238) and deleteBackup() (line ~204). (why: Aligns sensitive/exfiltration-capable backup endpoints with the deliberate admin-only fencing already applied to restoreBackup()/uploadRestore(); closes the full-DB+uploads download path for staff.)

**Verification:**
- Log in as a 'staff' user, POST /admin/updates/backup with valid CSRF: after fix it should 403 (if admin-only) or be rate-limited after N calls.
- Confirm an 'admin' user can still create/download backups unchanged.
- Run tests/backup-restore.spec.js to ensure the admin happy-path round-trip still passes.
- Hammer POST /admin/updates/backup repeatedly as admin → expect 429 after the rate-limit window.

**Edge cases to preserve:**
- Update flow's internal pre-update backup (Updater::createBackup / performUpdate) must not be rate-limited or blocked — it runs server-side, not via the HTTP endpoint.
- restoreBackup/uploadRestore already admin-gated — do not double-add or remove that check.
- scope='db' path must still work for whatever role is permitted.
- CSRF validation order must remain (don't let rate-limit bypass CSRF).

**Latest fix attempt (fixrun_20260610T055559Zae5b41):** fixed and verified

#### F034 — addslashes() is used to escape credential values written into a MySQL option file (.cnf), but addslashes() is not the correct escaping for MySQL option-file format — a backslash in DB_PASS or DB_USER (e.g. 'pa\ss') becomes a double backslash which MySQL interprets differently, causing authentication failure or value injection if the password contains special sequences.

**File:** `app/Support/BackupManager.php:621-629`
**Score:** 45
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-06-10T07:46:39Z — walkthrough: F034 — picked option A (Replace addslashes() with a cnfEscape() helper for MySQL option-file values)
_Promoted from disposition=`uncertain` / actionability=`auto_fixable` / score_phase4=45_
**Fix direction:** Add `private function cnfEscape(string $v): string { return str_replace(['\\', '"'], ['\\\\', '\\"'], $v); }` to app/Support/BackupManager.php (escape backslash FIRST, then double-quote). Replace every addslashes($cfg[...]) call in the .cnf writing block of importViaCli() (the [client] user/password/socket-or-host lines) with $this->cnfEscape(...). Single quotes must pass through verbatim — that is the bug being fixed. Do NOT change the fallback importer (importViaPhp), do NOT touch openImportConnection, do NOT alter importViaCli return paths.

**Latest fix attempt (fixrun_20260610T114937Z0f7a11):** fixed and verified

#### F036 — exec() output ($out) contains the merged stderr of the mysql CLI invocation (2>&1), which may include the full error message and is silently discarded — no logging occurs, so a failed CLI import is completely invisible in the audit trail.

**File:** `app/Support/BackupManager.php:637`
**Score:** 45
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-06-10T07:46:39Z — walkthrough: F036 — picked option A (Log stderr via SecureLogger::warning before falling back)
_Promoted from disposition=`uncertain` / actionability=`auto_fixable` / score_phase4=45_
**Fix direction:** In app/Support/BackupManager.php importViaCli(), before the cleanup/@unlink($cnf) on the failure path add: `if ($code !== 0) { SecureLogger::warning('[BackupManager] mysql CLI restore failed', ['exit_code' => $code, 'stderr_tail' => mb_substr(trim(implode("\n", $out)), -900)]); }`. Use context key 'stderr_tail' NOT 'stderr_key' (avoid key-name redaction match); trim to -900 not -2000 (SecureLogger::sanitizeString caps at 1000 chars); do NOT pass $cmd or the $cnf path in the context (avoid leaking tempdir/sqlPath). Import App\Support\SecureLogger if not already used in the file.

**Latest fix attempt (fixrun_20260610T114937Z0f7a11):** fixed and verified

#### F038 — The 2 GB upload size cap is enforced after the file has already been fully received by PHP (moveTo has completed) — an attacker can upload a 2 GB+ file and exhaust disk space before the check triggers.

**File:** `app/Support/BackupManager.php:243-260`
**Score:** 40
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-06-10T10:03:14Z — walkthrough: F038 — picked option A (Reject oversized upload in controller before moveTo)
_Promoted from disposition=`?` / actionability=`auto_fixable` / score_phase4=None_
**Fix direction:** In app/Controllers/UpdateController.php uploadRestore(), immediately after the UPLOAD_ERR_OK guard and BEFORE $upload->moveTo(...): `$earlySize = $upload->getSize(); if ($earlySize !== null && $earlySize > \App\Support\BackupManager::MAX_UPLOAD_BYTES) { return $this->jsonResponse($response, ['error' => __('File di backup troppo grande')], 400); }`. Make MAX_UPLOAD_BYTES public in BackupManager if currently private. Keep the existing post-move check in restoreFromUploadedZip as a backstop (do NOT remove it). Check whether __('File di backup troppo grande') already exists in the locales; if not, add to all 4.

**Latest fix attempt (fixrun_20260610T114937Z0f7a11):** fixed and verified

#### F039 — The manifest validation only checks that the 'app' key equals 'Pinakes' — an attacker who can supply an uploaded ZIP (admin-only, but still) could craft a manifest with app=Pinakes and a crafted database.sql containing arbitrary DROP/CREATE/INSERT statements, and the whole dump will be imported; the only integrity check is the SHA-256 which is also attacker-controlled in the manifest.

**File:** `app/Support/BackupManager.php:302-303`
**Score:** 35
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-06-10T10:03:14Z — walkthrough: F039 — picked option B (Trust-boundary warning copy in the upload-restore confirm dialog)
_Promoted from disposition=`?` / actionability=`auto_fixable` / score_phase4=None_
**Fix direction:** In app/Views/admin/updates.php, in the uploadRestoreFile() confirmation Swal (the 'Ripristinare da questo file?' dialog), add to the dialog text/html the advisory __('Ripristina solo archivi creati da questa istanza o di cui ti fidi completamente: il contenuto del database verrà sostituito con quello dell\'archivio.'). Add the key to all 4 locale JSONs, json_encode with JSON_HEX_TAG. Do NOT add HMAC logic or any server-side change (deferred to a future minor). Do NOT touch the stored-backup restore dialog.

**Latest fix attempt (fixrun_20260610T114937Z0f7a11):** fixed and verified

#### F041 — importViaPhp loads the entire SQL dump into memory via file_get_contents — for a large database this could exhaust PHP memory and kill the process mid-restore, leaving the database in a destroyed half-imported state with no rollback possible (DDL is auto-committed).

**File:** `app/Support/BackupManager.php:679-716`
**Score:** 45
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-06-10T07:46:39Z — walkthrough: F041 — picked option E (Combine pre-flight size guard with format-aware line-batch streaming)
_Promoted from disposition=`uncertain` / actionability=`manual` / score_phase4=45_
**Fix direction:** Two-part change in app/Support/BackupManager.php importViaPhp(): (1) SIZE GUARD first — before reading the dump, parse ini_get('memory_limit') ('-1' means unlimited → skip guard; convert 128M/1G shorthand to bytes) and `if (filesize($sqlPath) > $memoryBudget * 0.6) throw new \RuntimeException(...)` with a clear message, BEFORE any DDL runs so the existing DB is untouched. (2) STREAMING — replace file_get_contents + multi_query($main) with an fopen/fgets line-batch reader: accumulate lines into a batch string, flush via multi_query() every ~200 lines and at the '-- PINAKES_TRIGGERS' sentinel (keep the trigger-block split logic unchanged). This relies on the dump format invariant: dumpDatabaseTo() emits one SQL statement per physical line because real_escape_string escapes \n/\r — add a comment stating 'This parser assumes one statement per physical line as guaranteed by dumpDatabaseTo(); do NOT change dumpDatabaseTo() escaping without updating this importer.' Do NOT build a quote-state SQL parser. Do NOT remove the size guard after streaming works. Keep the SHA-256 manifest verification untouched. New user-facing error strings need __() + all 4 locale JSONs.

**Latest fix attempt (fixrun_20260610T114937Z0f7a11):** fixed and verified

</details>

## Polish — below threshold, clustered (3)

Below-gate findings (score < 45) that cluster in the same area — not worth surfacing individually, but dense enough that a human pass may catch something the pipeline filtered out.

| # | Score | File | Rough location | Concern |
|---|-------|------|----------------|---------|
| F022 | 22 | `app/Support/BackupManager.php` | L221-233 | The docstring for restoreFromBackup() says 'Restore a stored backup (DB + files)' but a db-scope ZIP (scope=db, no files/* entries) is accepted and will restore only the database while the comment implies both are always restored, misleading callers about the partial-restore behaviour. |
| F008 | 18 | `app/Support/BackupManager.php` | L254-256 | Probe-close guard relies on probe->filename being non-empty to decide whether to close after a failed open/locateName chain; logically correct today but fragile if ZipArchive ever returns an empty filename for an open archive. |
| F040 | 30 | `app/Support/BackupManager.php` | L254-260 | restoreFromUploadedZip opens the ZIP archive (probe.open) to validate it, then closes it and re-opens later in restoreZip — the TOCTOU window between the two opens means a symlink or a concurrent file replacement could substitute a different (malicious) archive between validation and import. |

## Fix runs

### Run `fixrun_20260610T055559Zae5b41` — 2026-06-10T06:21:44Z

- Outcomes: 5 fixed and verified
- Commits: `80ae16e`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F011 | FG-1 | ✓ fixed and verified |  |
| F016 | FG-1 | ✓ fixed and verified |  |
| F019 | FG-1 | ✓ fixed and verified |  |
| F023 | FG-1 | ✓ fixed and verified |  |
| F033 | FG-1 | ✓ fixed and verified |  |

### Run `fixrun_20260610T114937Z0f7a11` — 2026-06-10T12:18:00Z

- Outcomes: 14 fixed and verified
- Commits: `889ce06`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F013 | FG-1 | ✓ fixed and verified |  |
| F014 | FG-1 | ✓ fixed and verified |  |
| F015 | FG-1 | ✓ fixed and verified |  |
| F024 | FG-2 | ✓ fixed and verified |  |
| F025 | FG-2 | ✓ fixed and verified |  |
| F027 | FG-2 | ✓ fixed and verified |  |
| F028 | FG-2 | ✓ fixed and verified |  |
| F030 | FG-2 | ✓ fixed and verified |  |
| F032 | FG-2 | ✓ fixed and verified |  |
| F034 | FG-1 | ✓ fixed and verified |  |
| F036 | FG-1 | ✓ fixed and verified |  |
| F038 | FG-1 | ✓ fixed and verified |  |
| F039 | FG-2 | ✓ fixed and verified |  |
| F041 | FG-1 | ✓ fixed and verified |  |

fabiodalez-dev:
### Auto-recommendation acceptance

`rev_20260609T200240Z715548` · run_id=fixrun_20260610T055559Zae5b41 · choice=apply_all · threshold=50 · reviewer=auto-rec/fabiodalezbackup@gmail.com · ts=2026-06-10T06:02:19Z

4 auto-rec finding(s) eligible at threshold ≥ 50. Of those, **4** auto-promoted via batch, **0** promoted with an edited or alternative hint, **0** skipped per-finding.

#### Promoted (batch)

- **F011** — In restoreZip() the irreversible DB import (importDatabase, step 4) and the post-import promoteStagedFiles() (step 5) share one catch: a throw from promoteStagedFiles AFTER the DB has already been replaced returns success=false with a generic error, so the user is told the restore failed even though the database was already overwritten — they may retry or restore the safety backup over a partially-applied state.
  - **Hint:** `In `app/Support/BackupManager.php`, add a `$dbImported = false;` flag before the try block and set it to `true` immediately after line 324 (`$this->importDatabase(...)`). In the catch at line 335, branch on `$dbImported`: when `true`, return `['success'=>false,'partial'=>true,'restored_phase'=>'database','safety_backup'=>$safetyName,'error'=>$e->getMessage()]` instead of the plain failure shape. In `app/Controllers/UpdateController.php` lines 286-293 and 334-341, forward `partial` and `restored_phase` from `$result` into the JSON response. In `app/Views/admin/updates.php` `runRestore()` (~line 1161), add a `data.partial` branch before the generic error path that fires a `warning`-icon SweetAlert instructing the user not to retry and to restore only the safety backup if needed.`
- **F016** — json_encode() call emitting PHP data into a JS context does not pass JSON_HEX_TAG, violating the project view-escaping convention that requires JSON_HEX_TAG for every PHP-to-JS value in views.
  - **Hint:** `In `app/Views/admin/updates.php` line 563, change `json_encode(($_SESSION['user']['tipo_utente'] ?? '') === 'admin')` to `json_encode(($_SESSION['user']['tipo_utente'] ?? '') === 'admin', JSON_HEX_TAG)` to comply with the project view-escaping convention (JSON_HEX_TAG required on every PHP-to-JS json_encode in views).`
- **F023** — The 'Carica e Ripristina' button has no client-side file-type or size validation before showing the confirmation dialog, so the user gets a loading spinner and a server-side error after waiting through a potentially large upload, with no indication of the 2 GB hard cap or that only .zip files are accepted.
  - **Hint:** `In `app/Views/admin/updates.php`, inside `uploadRestoreFile()` (~line 1116), add client-side validation before the SweetAlert confirmation: check `!file.name.toLowerCase().endsWith('.zip')` and `file.size > 2 * 1024 * 1024 * 1024` and fire a `Swal.fire({ icon:'warning', ... })` with an explicit message (e.g. "Solo file .zip fino a 2 GB sono accettati") returning early if either check fails. Also update the `<p>` hint text (~line 262) to mention the 2 GB cap and `.zip`-only restriction.`
- **F033** — createBackup endpoint has no admin role re-check — any authenticated staff user can trigger a potentially large full-backup (DB + all uploads), creating a resource-exhaustion risk, since the backup ZIP is stored under storage/backups and later downloadable.
  - **Hint:** `Two-part fix: (1) In `app/Routes/web.php` lines 3042-3046, append `->add(new \App\Middleware\RateLimitMiddleware(5, 60))` to the `POST /admin/updates/backup` route, matching the rate-limit pattern on other update routes at line 3022. (2) In `app/Controllers/UpdateController.php`, add the same admin re-check already present at lines 301-303 to `createBackup()` (line ~124), `downloadBackup()` (~line 238), and `deleteBackup()` (~line 204) so all three endpoints are gated to admin-only, consistent with `restoreBackup()`/`uploadRestore()`.`

---

Auto-recommendation acceptance: append-only audit. Each fix run posts a fresh entry. Promoted findings are now `confirmed_mechanical` with `human_confirmation` set; the fix phase dispatches them next.

fabiodalez-dev:
### Walkthrough decisions

`rev_20260609T200240Z715548` · scope=**qualifying** · score_floor=30 · reviewer=fabiodalezbackup@gmail.com · ts=2026-06-10T07:47:33Z

Walking the **Qualifying** scope: of **5** non-auto-eligible finding(s), **5 promoted**, **0 skipped**, **0 stopped**.

Promoted findings will be picked up by the next fix run via the `human_confirmation` bypass, regardless of its score threshold.

---

#### Promoted

- **F014** — extractFilesSafely() streams each ZIP entry to disk with no cap on cumulative decompressed size; an admin-uploaded restore ZIP is only size-capped on its compressed form, so a decompression-bomb archive can exhaust disk during restore and leave a half-promoted filesystem. · option **A** (Add MAX_RESTORE_DECOMPRESSED_BYTES constant and track cumulative bytes written)
  - **Fix hint:** `Add `private const MAX_RESTORE_DECOMPRESSED_BYTES = 4 * 1024 * 1024 * 1024;` after MAX_UPLOAD_BYTES in app/Support/BackupManager.php. In extractFilesSafely(), initialise `$totalDecompressed = 0;` before the loop, replace `stream_copy_to_stream($stream, $out);` with `$written = stream_copy_to_stream($stream, $out); $totalDecompressed += $written;` and immediately after add `if ($totalDecompressed > self::MAX_RESTORE_DECOMPRESSED_BYTES) { fclose($stream); fclose($out); throw new \RuntimeException(__('Archivio di backup non valido: dimensione decompressa eccessiva')); }`. New user-facing string must be added to all 4 locale JSONs per project rule. Do NOT add per-entry statIndex pre-flight, do NOT pre-scan the archive — cumulative counter only.`
  - **Prior:** disposition=uncertain · score=45
- **F025** — The restore confirmation dialog uses a single-step SweetAlert with the confirm button as the default focused element; for an operation that overwrites the entire database and all uploaded files, the focused element on dialog open should be the safe action ('Annulla'), not the destructive 'Ripristina' button, to prevent an accidental Enter keypress from triggering the restore. · option **A** (Add focusCancel: true to both destructive restore dialogs)
  - **Fix hint:** `In app/Views/admin/updates.php, inside the Swal.fire confirmation call in restoreBackup() (the one with confirmButtonColor '#d97706') add `focusCancel: true,`; repeat identically for the uploadRestoreFile() confirmation Swal.fire. Do NOT change confirmButtonColor, button text, icon, or any other Swal option. Touch only these two dialog objects. Locate dialogs by content, not line numbers (file shifted after recent fixes).`
  - **Prior:** disposition=uncertain · score=58
- **F034** — addslashes() is used to escape credential values written into a MySQL option file (.cnf), but addslashes() is not the correct escaping for MySQL option-file format — a backslash in DB_PASS or DB_USER (e.g. 'pa\ss') becomes a double backslash which MySQL interprets differently, causing authentication failure or value injection if the password contains special sequences. · option **A** (Replace addslashes() with a cnfEscape() helper for MySQL option-file values)
  - **Fix hint:** `Add `private function cnfEscape(string $v): string { return str_replace(['\\', '"'], ['\\\\', '\\"'], $v); }` to app/Support/BackupManager.php (escape backslash FIRST, then double-quote). Replace every addslashes($cfg[...]) call in the .cnf writing block of importViaCli() (the [client] user/password/socket-or-host lines) with $this->cnfEscape(...). Single quotes must pass through verbatim — that is the bug being fixed. Do NOT change the fallback importer (importViaPhp), do NOT touch openImportConnection, do NOT alter importViaCli return paths.`
  - **Prior:** disposition=uncertain · score=45
- **F036** — exec() output ($out) contains the merged stderr of the mysql CLI invocation (2>&1), which may include the full error message and is silently discarded — no logging occurs, so a failed CLI import is completely invisible in the audit trail. · option **A** (Log stderr via SecureLogger::warning before falling back)
  - **Fix hint:** `In app/Support/BackupManager.php importViaCli(), before the cleanup/@unlink($cnf) on the failure path add: `if ($code !== 0) { SecureLogger::warning('[BackupManager] mysql CLI restore failed', ['exit_code' => $code, 'stderr_tail' => mb_substr(trim(implode("\n", $out)), -900)]); }`. Use context key 'stderr_tail' NOT 'stderr_key' (avoid key-name redaction match); trim to -900 not -2000 (SecureLogger::sanitizeString caps at 1000 chars); do NOT pass $cmd or the $cnf path in the context (avoid leaking tempdir/sqlPath). Import App\Support\SecureLogger if not already used in the file.`
  - **Prior:** disposition=uncertain · score=45
- **F041** — importViaPhp loads the entire SQL dump into memory via file_get_contents — for a large database this could exhaust PHP memory and kill the process mid-restore, leaving the database in a destroyed half-imported state with no rollback possible (DDL is auto-committed). · option **E** (Combine pre-flight size guard with format-aware line-batch streaming)
  - **Fix hint:** `Two-part change in app/Support/BackupManager.php importViaPhp(): (1) SIZE GUARD first — before reading the dump, parse ini_get('memory_limit') ('-1' means unlimited → skip guard; convert 128M/1G shorthand to bytes) and `if (filesize($sqlPath) > $memoryBudget * 0.6) throw new \RuntimeException(...)` with a clear message, BEFORE any DDL runs so the existing DB is untouched. (2) STREAMING — replace file_get_contents + multi_query($main) with an fopen/fgets line-batch reader: accumulate lines into a batch string, flush via multi_query() every ~200 lines and at the '-- PINAKES_TRIGGERS' sentinel (keep the trigger-block split logic unchanged). This relies on the dump format invariant: dumpDatabaseTo() emits one SQL statement per physical line because real_escape_string escapes \n/\r — add a comment stating 'This parser assumes one statement per physical line as guaranteed by dumpDatabaseTo(); do NOT change dumpDatabaseTo() escaping without updating this importer.' Do NOT build a quote-state SQL parser. Do NOT remove the size guard after streaming works. Keep the SHA-256 manifest verification untouched. New user-facing error strings need __() + all 4 locale JSONs.`
  - **Prior:** disposition=uncertain · score=45

---

Decisions log: this comment is append-only audit — it is never edited in place. Each walkthrough run posts a fresh entry. Current state: see the main review comment and artifact.md.

fabiodalez-dev:
### Walkthrough decisions

`rev_20260609T200240Z715548` · scope=**full** · score_floor=30 · reviewer=fabiodalezbackup@gmail.com · ts=2026-06-10T10:05:31Z

Walking the **Full skip set** scope: of **10** non-auto-eligible finding(s), **9 promoted**, **1 skipped**.

Promoted findings will be picked up by the next fix run via the `human_confirmation` bypass, regardless of its score threshold.

---

#### Promoted

- **F013** — dumpDatabaseTo() serializes every column value via real_escape_string()+single-quote regardless of type; binary/BLOB column bytes are emitted as a quoted text l · option **C** (Document the no-BLOB invariant in code comment)
  - **Prior:** disposition=below_gate · score=null
- **F015** — importViaCli() treats any non-zero exit as a clean retry signal and falls through to the PHP importer, but `mysql < dump` aborts on the first failing statement  · option **A** (Strengthen the docblock to document trigger idempotency)
  - **Prior:** disposition=below_gate · score=null
- **F024** — The restore progress modal shows only an indeterminate spinner with generic text — for a potentially multi-GB upload-then-restore there is no upload progress ba · option **B+A** (XHR upload progress bar + elapsed-time counter)
  - **Prior:** disposition=below_gate · score=null
- **F027** — On a successful restore the success dialog shows only the safety-backup filename with no guidance on what to do next — the page state is now inconsistent (DB/se · option **B** (Success dialog confirm button that reloads on click)
  - **Prior:** disposition=below_gate · score=null
- **F028** — Legacy directory backups (update_* folders, DB-only) show a 'Ripristina' button only for .zip backups, silently rendering them non-restorable with no explanator · option **A** (Muted 'Formato legacy' badge with tooltip)
  - **Prior:** disposition=below_gate · score=null
- **F030** — The manual-backup scope selector ('Completo' vs 'Solo database') is shown to all logged-in users (not gated to admin), so a staff user who cannot trigger a rest · option **A** (PHP-side admin gate on the backup-creation card)
  - **Prior:** disposition=below_gate · score=null
- **F032** — When the restore fails with a non-JSON response (PHP fatal, gateway timeout) the displayed error is 'Errore durante il ripristino (HTTP 502)' with no guidance o · option **A** (Non-JSON restore error dialog with recovery guidance)
  - **Prior:** disposition=below_gate · score=null
- **F038** — The 2 GB upload size cap is enforced after the file has already been fully received by PHP (moveTo has completed) — an attacker can upload a 2 GB+ file and exha · option **A** (Reject oversized upload in controller before moveTo)
  - **Prior:** disposition=below_gate · score=null
- **F039** — The manifest validation only checks that the 'app' key equals 'Pinakes' — an attacker who can supply an uploaded ZIP (admin-only, but still) could craft a manif · option **B** (Trust-boundary warning copy in upload-restore dialog)
  - **Prior:** disposition=below_gate · score=null

#### Skipped

- **F040** — TOCTOU between ZIP probe and re-open
  - Reviewer skipped: theoretical — random destination path + downstream SHA-256 verification close the practical risk.

---

Decisions log: append-only audit; each walkthrough run posts a fresh entry. Current state: see the main review comment and artifact.md.

fabiodalez-dev:
### Code review

**Branch:** `feature/162-complete-backup` → `main`
**PR:** #167 (open)
**Review ID:** `rev_20260610T132541Z6da1b0`
**Sub-agent tokens:** 1,871,967 across 15 invocations

Found 18 findings across all lanes:
- Deep lane (correctness/security): 3 resolved, 5 uncertain

---

### ✓ Auto-fixable (3)

| # | Score | Impact | File | Issue | Status |
|---|-------|--------|------|-------|--------|
| F002 | 60 | correctness | `app/Support/BackupManager.php:1060-1065` | The zip-bomb cumulative-size guard checks `$totalDecompressed > MAX_RESTORE_DECOMPRESSED_BYTES` only AFTER `stream_copy_to_stream` has fully written the current entry to disk; a single entry larger than 4 GB would be written in its entirety before the check fires. | ✓ fixed and verified (`f89255b`) |
| F003 | 75 | correctness | `app/Support/BackupManager.php:1060-1068` | `stream_copy_to_stream` can return `false` on I/O failure, but the return value is not checked; a failed copy silently adds 0 to `$totalDecompressed` and leaves a zero-byte or truncated file in the staging directory, which is then promoted over the live upload root. | ✓ fixed and verified (`f89255b`) |
| F005 | 75 | correctness | `app/Support/BackupManager.php:286-345` | Restore drops and recreates every table on a live site with neither maintenance mode nor a concurrency lock: Updater::performUpdate takes a lock file (flock) and calls enableMaintenanceMode() before touching anything (Updater.php:1595-1599), but the new restoreBackup/uploadRestore paths run importDatabase (DROP TABLE ... CREATE ... INSERT) with no lock and no maintenance flag, so concurrent user requests hit missing tables mid-restore and two concurrent restores (double-click, two admins) can interleave DDL and corrupt the database. | ✓ fixed and verified (`f89255b`) |

**Cross-cutting group G1:** F002 + F003 — Both findings are defects of the same two stream_copy_to_stream call sites in app/Support/BackupManager.php (extractFilesSafely line 1060 and streamEntryToFile line 384), and their independent fix proposals would rewrite the same lines — they must be fixed as one change. Replace both unbounded, unchecked stream_copy_to_stream calls with a single private chunked-copy helper (e.g. copyEntryStream($stream, $out, &$totalDecompressed, $expectedSize, $relative, $target)) that loops fread($stream, 1 MiB)/fwrite: (1) F002 — increment the shared cumulative $totalDecompressed budget per chunk and throw the existing RuntimeException(__('Archivio di backup non valido: dimensione decompressa eccessiva')) the moment it exceeds MAX_RESTORE_DECOMPRESSED_BYTES, bounding bytes-on-disk to cap + one chunk instead of one full entry; keep the budget cumulative across entries (do NOT reset per entry) and apply the same budget to the database.sql path in streamEntryToFile, which today has no cap at all. (2) F003 — treat fread === false and short fwrite as hard I/O failures throwing RuntimeException(__('Errore durante il ripristino dei file') . ': ' . $relative) per the documented fail-loud contract at lines 1028-1031, and after the loop compare total bytes written against ZipArchive::statIndex($i)['size'] so a short copy (disk-full returns a short count, not false) also throws — for both extractFilesSafely entries and streamEntryToFile's database.sql (whose SHA-256 check is bypassed when the manifest sha is empty). On every throw path: fclose both handles and @unlink the partial $target/$dest. Keep statIndex size only as advisory/verification (forgeable), with the chunked runtime budget authoritative for F002. Preserve the security `continue` skips (zip-slip/symlink/out-of-root, lines 1014-1050) as silent skips, zero-byte entries terminating via feof, and the staging-dir/temp-SQL cleanup in restoreZip's catch. Verify with: grep stream_copy_to_stream returns no unchecked sites; an adversarial-zip test (high-ratio entry past a test-lowered cap → 'dimensione decompressa eccessiva'; corrupted/truncated entry → restore fails, live files untouched); and the existing /tmp/run-e2e.sh tests/backup-restore.spec.js round-trip still passes.

<details><summary>Details and fix proposals</summary>

#### F002 — The zip-bomb cumulative-size guard checks `$totalDecompressed > MAX_RESTORE_DECOMPRESSED_BYTES` only AFTER `stream_copy_to_stream` has fully written the current entry to disk; a single entry larger than 4 GB would be written in its entirety before the check fires.

**File:** `app/Support/BackupManager.php:1060-1065`
**Score:** 60 (moderate)

**Evidence:**
- app/Support/BackupManager.php:1060-1066 — `$written = stream_copy_to_stream($stream, $out); $totalDecompressed += (int) $written; if ($totalDecompressed > self::MAX_RESTORE_DECOMPRESSED_BYTES) { ... throw }` — the copy is unbounded (no maxlength argument) and the cap is evaluated only after the entire entry is on disk
- app/Support/BackupManager.php:43 — `MAX_RESTORE_DECOMPRESSED_BYTES = 4 * 1024 * 1024 * 1024`; lines 1008-1009 comment states the guard's purpose is 'against a zip bomb ... the expanded form is not [capped]' — the within-entry overshoot defeats exactly that stated purpose, so this is an incomplete fix, not a documented tradeoff
- git show 889ce06a — the guard was added in the round-1 fix commit ('address walkthrough-promoted review findings'); the candidate identifies a gap in that fix, which is new evidence, not a re-flag
- Reproduction: craft a ZIP containing manifest.json, database.sql, and one entry `files/public/uploads/copertine/bomb.bin` that is ~5 MB compressed but decompresses to e.g. 50 GB of zeros (DEFLATE ratio up to ~1032:1, so compressed size stays far below MAX_UPLOAD_BYTES=2GB at line 255 and UpdateController.php:336 early check); admin uploads it via uploadRestore → extractFilesSafely streams the full 50 GB (or until disk-full) into the staging dir under sys_get_temp_dir() BEFORE line 1062 throws
- Severity limiters verified: endpoint is admin-only + CSRF (app/Controllers/UpdateController.php:317-323 and 280-287); staging dir is removed in restoreZip's catch (lines 351-353); so impact is transient tmp/disk exhaustion (DoS window during extraction), not persistent compromise
- Parallel path: app/Support/BackupManager.php:373-387 streamEntryToFile() copies the database.sql entry with stream_copy_to_stream and NO size cap at all, and runs at line 318 before the hash check inside importDatabase (line 332) — same class, must be covered by the fix
- No adversarial-zip test exists: tests/backup-restore.spec.js covers round-trip only; no test references zip-bomb/decompression limits

**Approach:** Replace the two unbounded stream_copy_to_stream calls with a chunked copy helper that increments the shared decompressed-bytes budget per chunk (e.g. 1 MiB fread/fwrite) and throws the existing 'dimensione decompressa eccessiva' RuntimeException the moment the budget is exceeded, unlinking the partial target. Optionally pre-check ZipArchive::statIndex()['size'] per entry as a cheap fast-fail, but keep the chunked check as the authoritative guard since central-directory sizes can be forged. Give database.sql the same budget (or its own cap) in streamEntryToFile. Also treat a false/short return from the copy as a hard failure per the line 1028-1031 contract.

**Files to modify:**
- `app/Support/BackupManager.php` — In extractFilesSafely (lines 1051-1069): replace `stream_copy_to_stream($stream, $out)` with a loop `while (!feof($stream)) { $chunk = fread($stream, 1048576); if ($chunk === false) { cleanup; throw RuntimeException(__('Errore durante il ripristino dei file').': '.$relative); } $totalDecompressed += strlen($chunk); if ($totalDecompressed > self::MAX_RESTORE_DECOMPRESSED_BYTES) { fclose both; @unlink($target); throw RuntimeException(__('Archivio di backup non valido: dimensione decompressa eccessiva')); } if (fwrite($out, $chunk) !== strlen($chunk)) { cleanup; throw I/O RuntimeException; } }` (why: Aborts mid-entry, bounding total bytes written to MAX_RESTORE_DECOMPRESSED_BYTES + one chunk, instead of one full (unbounded) entry; also fixes the masked false-return of stream_copy_to_stream)
- `app/Support/BackupManager.php` — In streamEntryToFile (lines 373-387): apply the same chunked loop with the same MAX_RESTORE_DECOMPRESSED_BYTES cap (pass/share the budget from restoreZip, or apply the cap standalone), throwing on cap-exceeded and on fread/fwrite failure; unlink the partial $dest on abort (why: database.sql is the same untrusted-zip surface with NO cap at all today and is written before hash verification — fixing only extractFilesSafely leaves the worse half of the class open (partial fixes resurface next review))

**Verification:**
- grep -n 'stream_copy_to_stream' app/Support/BackupManager.php — must return no hits in extractFilesSafely or streamEntryToFile (or only hits inside the new budget-checked helper)
- Read extractFilesSafely and streamEntryToFile: confirm the budget check executes inside the read loop (per chunk), not after the copy, and that fread===false / short fwrite throw
- Unit/E2E: build a ZIP with valid manifest.json + database.sql + one files/ entry of ~50 MB compressed zeros declaring/decompressing past a test-lowered cap; assert restore fails with 'dimensione decompressa eccessiva' and that bytes written to the staging dir never exceeded cap + chunk size
- Run existing round-trip test: /tmp/run-e2e.sh tests/backup-restore.spec.js --config=tests/playwright.config.js --workers=1 — legitimate backup/restore must still pass

**Edge cases to preserve:**
- Legitimate backups with total decompressed size under 4 GB (including many entries summing close to the cap) must restore successfully — the budget is cumulative across entries, do not reset it per entry
- Security skips (zip-slip, symlink, out-of-root at lines 1014-1050) must keep using `continue`, while real I/O failures keep throwing — do not conflate the two (documented contract at lines 1028-1031)
- Staging dir cleanup on throw (restoreZip catch lines 351-353) and temp SQL cleanup (lines 347-350) must still run; the new mid-entry throw paths must close both stream handles before throwing
- statIndex 'size' can be forged in a crafted ZIP — the chunked runtime check must remain authoritative even if a declared-size fast-fail is added
- Empty entries / zero-byte files: fread loop must terminate via feof, not via byte-count assumptions

**Latest fix attempt (fixrun_20260610T140427Z39d6f2):** fixed and verified

#### F003 — `stream_copy_to_stream` can return `false` on I/O failure, but the return value is not checked; a failed copy silently adds 0 to `$totalDecompressed` and leaves a zero-byte or truncated file in the staging directory, which is then promoted over the live upload root.

**File:** `app/Support/BackupManager.php:1060-1068`
**Score:** 75 (strong)

**Evidence:**
- app/Support/BackupManager.php:1060-1061 — `$written = stream_copy_to_stream($stream, $out); $totalDecompressed += (int) $written;` — return value never checked against false or against the entry's expected uncompressed size; `(int) false === 0`, so a failed copy contributes nothing to the zip-bomb counter and does not throw.
- app/Support/BackupManager.php:1028-1031 — the PR's own invariant comment: 'Real I/O failures (mkdir/realpath/getStream/fopen) must FAIL the restore, not be silently skipped'. stream_copy_to_stream failure is exactly such an I/O failure and is absent from the enforced list — this is NOT a documented tradeoff; it contradicts the documented intent.
- git log -L 1055,1070: commit 715b67f7 (round-1 fixes) converted the fopen failure from `continue` to `throw`, establishing the fail-on-I/O-error invariant, but left stream_copy_to_stream unchecked; commit 889ce06a (round-2) captured $written only for the zip-bomb cap, with the (int) cast silently mapping false to 0.
- app/Support/BackupManager.php:340-345 — promoteStagedFiles($stagingDir) runs AFTER importDatabase(); a truncated/zero-byte staged file is moved over the live upload root and restoreZip() returns ['success' => true], so the corruption is reported as a successful restore.
- Reproduction: a backup zip whose files/public/uploads/copertine/x.jpg entry has a corrupted deflate stream (or disk-full on sys_get_temp_dir() mid-staging) → getStream succeeds, stream_copy_to_stream returns false or a short count → no throw → truncated file staged → promoted over the good live cover → admin sees 'restore successful'. PHP's stream_copy_to_stream returns a short byte count (not false) on mid-stream write failure, so a bare `=== false` check is insufficient — the written count must be compared to $zip->statIndex($i)['size'].
- Parallel path: app/Support/BackupManager.php:384 (streamEntryToFile) has the identical unchecked stream_copy_to_stream for database.sql. Mitigated by the SHA-256 verify in importDatabase() (lines 612-617) ONLY when the manifest sha is non-empty; line 112 stores `hash_file('sha256', $sqlTmp) ?: ''`, so a backup created when hash_file fails carries an empty sha and a truncated dump would be imported (dump begins with DROP TABLE IF EXISTS — catastrophic).

**Approach:** Enforce the existing fail-on-I/O-error invariant on the copy step itself, at both call sites. In extractFilesSafely(), fetch the entry's expected uncompressed size via $zip->statIndex($i) before copying, then after stream_copy_to_stream require ($written !== false && $written === $expectedSize) — on violation, fclose both handles and throw the same \RuntimeException(__('Errore durante il ripristino dei file') . ': ' . $relative) already used for the getStream/fopen failures at lines 1053/1058 (in-repo precedent). Comparing to the stat size (not just !== false) is essential because PHP returns a short byte count, not false, on mid-stream write failure (disk full). Keep feeding the actual written bytes into $totalDecompressed for the zip-bomb cap. In streamEntryToFile(), apply the same check (false → throw; size comparison optional there since the entry size is known via statName) so a truncated database.sql can never reach importDatabase() even when the manifest sha is empty. Update the lines 1028-1031 comment to include the copy step in the enforced list. Optionally check fclose($out) === false on the data handle (buffered flush failure) — same failure class.

**Files to modify:**
- `app/Support/BackupManager.php` — extractFilesSafely(): before the copy, $stat = $zip->statIndex($i); $expected = is_array($stat) ? (int) $stat['size'] : null. After $written = stream_copy_to_stream(...): if ($written === false || ($expected !== null && $written !== $expected)) { fclose($stream); fclose($out); throw new \RuntimeException(__('Errore durante il ripristino dei file') . ': ' . $relative); }. Keep $totalDecompressed += (int) $written and the MAX_RESTORE_DECOMPRESSED_BYTES check unchanged. Extend the comment at lines 1028-1031 to list the copy step among enforced failures. (why: Primary defect: a false/short copy currently stages a zero-byte or truncated file that promoteStagedFiles() moves over the live upload root after the DB import, and restoreZip() reports success — silent data corruption. Throwing here lands in the existing catch, which correctly reports partial=false (staging happens before importDatabase, so the DB is untouched) and cleans the staging dir.)
- `app/Support/BackupManager.php` — streamEntryToFile(): replace the bare stream_copy_to_stream($in, $out) at line 384 with a checked version: if (stream_copy_to_stream($in, $out) === false) { fclose($in); fclose($out); throw new \RuntimeException(__('Errore durante il ripristino dei file') . ': ' . $entry); } (or compare against $zip->statName($entry)['size'] for short-write detection). (why: Same defect class on the database.sql path. The SHA-256 verify in importDatabase() covers it only when manifest database_sha256 is non-empty; line 112 (hash_file(...) ?: '') can legitimately produce an empty sha, in which case a truncated dump (nonzero size, so the filesize()===0 guard passes) would be imported — and the dump starts with DROP TABLE IF EXISTS, destroying tables. Partial fixes here resurface next review.)

**Verification:**
- grep -n 'stream_copy_to_stream' app/Support/BackupManager.php — both call sites (currently lines 384 and 1060) must have their return value checked (false and, in extractFilesSafely, short-write vs statIndex size) and throw on failure
- Read extractFilesSafely() and confirm the throw path closes BOTH $stream and $out before throwing (matching the existing zip-bomb throw at lines 1063-1064), and that $totalDecompressed still accumulates actual written bytes
- Confirm the security skips (symlink at 1024, out-of-root at 1048) remain `continue`, not throws
- Run the existing E2E restore round-trip test (54b7754a 'verify downloaded backup is actually restorable') to confirm the happy path still passes; optionally add a test that corrupts one files/ entry in a backup zip and asserts restore fails with success=false and the live file untouched

**Edge cases to preserve:**
- Legitimate zero-byte files inside the archive: statIndex size 0 with $written === 0 must NOT throw (size comparison, not truthiness, handles this)
- Zip-bomb guard semantics: MAX_RESTORE_DECOMPRESSED_BYTES check must still fire on cumulative actual bytes written
- Security `continue` paths (out-of-root, symlink target, traversal names) must stay silent skips per the documented design
- Failure during staging (pre-importDatabase) must still report a clean non-partial failure; the partial=true protocol applies only after $dbImported = true

**Latest fix attempt (fixrun_20260610T140427Z39d6f2):** fixed and verified

#### F005 — Restore drops and recreates every table on a live site with neither maintenance mode nor a concurrency lock: Updater::performUpdate takes a lock file (flock) and calls enableMaintenanceMode() before touching anything (Updater.php:1595-1599), but the new restoreBackup/uploadRestore paths run importDatabase (DROP TABLE ... CREATE ... INSERT) with no lock and no maintenance flag, so concurrent user requests hit missing tables mid-restore and two concurrent restores (double-click, two admins) can interleave DDL and corrupt the database.

**File:** `app/Support/BackupManager.php:286-345`
**Score:** 75 (strong)

**Evidence:**
- app/Support/BackupManager.php:286-368 — restoreZip() body contains no flock, no lock file, no maintenance-file write; grep -ni 'maintenance|lock|concurrent' over the file matches only an unrelated dump-transaction comment (line 481) and trigger-block parsing comments.
- app/Support/BackupManager.php:607-685 — importDatabase() runs the full DROP/CREATE/INSERT dump via mysql CLI (importViaCli) or PHP fallback; every table is dropped then recreated sequentially, so concurrent SELECTs on 'libri' etc. hit ER_NO_SUCH_TABLE for the duration of the import.
- app/Support/Updater.php:1575-1599 — in-repo precedent: performUpdate() acquires flock(LOCK_EX|LOCK_NB) on storage/cache/update.lock and calls enableMaintenanceMode() before any destructive step; Updater.php:1536-1551 registers a shutdown handler that removes both .maintenance and the lock on fatal error; Updater.php:3204-3237 enable/disable implementations.
- public/index.php:136-170 — the maintenance file is enforced for every request at the front controller (503 + Retry-After), with an allowlist ['/admin/updates', '/assets/', '/favicon.ico'] that already permits the restore endpoints (/admin/updates/backup/restore, /admin/updates/backup/restore-upload) and the admin polling UI to keep working during maintenance — so enabling maintenance from the restore path needs no allowlist change.
- app/Controllers/UpdateController.php:294 and :352 — both restore endpoints construct BackupManager and call restoreFromBackup/restoreFromUploadedZip directly; no lock or maintenance call in the controller either; routes wired at app/Routes/web.php:3082-3094 with only CsrfMiddleware + AdminAuthMiddleware.
- app/Views/admin/updates.php:1106-1230 — the only concurrency guard is a same-tab SweetAlert blocking modal during runRestore(); it does not prevent a second browser tab, a second admin session, or any concurrent public traffic.
- Not a documented tradeoff: restoreZip/importViaCli comments document DDL non-transactionality + safety backup (293-294), hash-verified import (600-606), staging-first extraction (320-323), and the partial-restore protocol (333-335, 354-364) — none mentions concurrency, locking, or maintenance mode. git log main..HEAD and fix commits 80ae16e5/889ce06a show no prior finding on this topic.
- Reproduction: admin POSTs /admin/updates/backup/restore on a live site; while importDatabase is mid-dump (seconds to minutes), any visitor request queries dropped tables → mysqli errors/500s with no 503 maintenance page; two admins (or one admin in two tabs) POSTing restore concurrently each pass CSRF+role checks and run interleaved DROP/CREATE/INSERT streams against the same schema, and each takes a safety backup of the other's half-imported state.

**Approach:** Mirror the Updater's existing protection inside BackupManager::restoreZip() so both restore endpoints are covered at the single convergence point: (1) acquire flock(LOCK_EX|LOCK_NB) on storage/cache/update.lock — sharing the updater's lock file (Updater.php:1508) serializes restore-vs-restore AND restore-vs-update; on contention return ['success'=>false, 'error'=>__('Un'altra operazione di manutenzione è già in corso. Riprova più tardi.')]. (2) Immediately after acquiring the lock, write storage/.maintenance with the same JSON shape the front controller expects ({message, time} — public/index.php:159-163), so all non-/admin/updates traffic gets a clean 503 instead of table-not-found errors; do this BEFORE the safety backup so the whole destructive window is quiesced. (3) Wrap the body in try/finally: finally removes .maintenance and releases/unlinks the lock on success, failure, AND the partial-restore path; add a register_shutdown_function safety net exactly like Updater.php:1536-1551 so a fatal error mid-import cannot leave the site locked (the 30-min staleness fallback at index.php:162-166 remains the last resort). (4) Also set set_time_limit(0) + ignore_user_abort(true) at the top of restoreZip, matching Updater.php:1553-1554, so the import cannot be aborted mid-DDL by PHP timeout or client disconnect. (5) Add the new error string to all 4 locale JSONs per the project's i18n BLOCKING rule. No change needed in index.php: the allowlist (line 149) already lets /admin/updates/backup/restore* and the polling UI through during maintenance.

**Files to modify:**
- `app/Support/BackupManager.php` — In restoreZip(): set_time_limit(0)/ignore_user_abort(true); open+flock storage/cache/update.lock (LOCK_EX|LOCK_NB, write PID) returning a friendly 'operation already in progress' error on contention; write storage/.maintenance ({message,time} JSON) before the safety backup; register_shutdown_function cleanup; finally-block that unlinks .maintenance and releases+unlinks the lock on every exit path including the dbImported/partial branch. (why: restoreZip is the single convergence point of both restore endpoints (restoreFromBackup and restoreFromUploadedZip), so protecting it covers the full class; sharing update.lock also closes the restore-during-update race. Precedent: Updater.php:1553-1554 (limits), 1575-1599 (lock+maintenance), 1536-1551 (shutdown cleanup), 3204-3237 (maintenance file format).)
- `locale/it_IT.json` — Add the new lock-contention message key (e.g. "Un'altra operazione di manutenzione è già in corso. Riprova più tardi."). (why: Project rule: every user-facing string wrapped in __() must land in all 4 locale JSONs in the same commit (feedback_ui_style_and_i18n_blocking).)
- `locale/en_US.json` — Add the same key with the English translation. (why: i18n BLOCKING rule — all 4 locales in the same commit.)
- `locale/de_DE.json` — Add the same key with the German translation. (why: i18n BLOCKING rule — all 4 locales in the same commit.)
- `locale/fr_FR.json` — Add the same key with the French translation. (why: i18n BLOCKING rule — all 4 locales in the same commit.)

**Verification:**
- grep -n 'flock' app/Support/BackupManager.php — must show LOCK_EX|LOCK_NB acquisition in restoreZip before createBackup('full') is called
- grep -n '.maintenance' app/Support/BackupManager.php — maintenance file written after lock acquisition and removed in the finally block; read the finally block to confirm it runs on the success, plain-failure, AND dbImported/partial paths
- grep -n 'register_shutdown_function' app/Support/BackupManager.php — shutdown safety net present, mirroring Updater.php:1536
- grep -n 'set_time_limit\|ignore_user_abort' app/Support/BackupManager.php — both set at the top of restoreZip
- grep the new error message key in all of locale/it_IT.json, en_US.json, de_DE.json, fr_FR.json
- Run E2E: /tmp/run-e2e.sh tests/backup-restore.spec.js --config=tests/playwright.config.js --workers=1 — full round-trip must still pass (the restore request itself must NOT be blocked by its own maintenance file thanks to the /admin/updates allowlist at public/index.php:149)
- Manual concurrency check: fire two simultaneous POSTs to /admin/updates/backup/restore — exactly one proceeds, the other gets the 'già in corso' JSON error; during the winner's import, a curl to the public catalog returns 503 with Retry-After, not a 500

**Edge cases to preserve:**
- Partial-restore protocol: when dbImported=true and file promotion fails, the response must still carry partial=true/restored_phase='database'/safety_backup — the finally cleanup must not swallow or reorder this return
- The maintenance allowlist (public/index.php:149) lets /admin/updates* through — the admin UI's elapsed-time polling and the restore POST itself must keep working during maintenance; do not write the maintenance file before the request's own routing if a different mechanism is chosen
- Updater::performUpdate at Updater.php:1839 calls BackupManager::createBackup() only — createBackup must NOT take the shared lock (the updater already holds it; locking there would deadlock the pre-update backup)
- 30-minute staleness auto-clear (index.php:162-166): a very long restore loses maintenance protection — same accepted limitation as the updater; consider refreshing the timestamp between phases but do not silently extend beyond the updater's behavior without flagging it
- Lock file storage/cache/ dir may not exist — mkdir like Updater.php:1570-1573
- safety_backup name must still be returned even when the lock-contention error path returns early (it is null there — keep the existing array shape)

**Latest fix attempt (fixrun_20260610T140427Z39d6f2):** fixed and verified

</details>

### ℹ Uncertain (5)

| # | Score | Impact | File | Issue |
|---|-------|--------|------|-------|
| F006 | 58 | correctness | `app/Support/BackupManager.php:373-387` | The decompression-bomb guard (MAX_RESTORE_DECOMPRESSED_BYTES) only counts files/* entries inside extractFilesSafely; streamEntryToFile copies the database.sql entry to disk with no decompressed-size cap at all, so a small uploaded ZIP whose database.sql entry decompresses to hundreds of GB exhausts the temp disk before the hash check ever runs, bypassing exactly the protection the PR added. |
| F008 | 50 | correctness | `app/Support/BackupManager.php:269-281` | restoreFromUploadedZip permanently moves the uploaded archive into storage/backups before the manifest app-identity check (which only happens later in restoreZip), so a ZIP that merely contains manifest.json + database.sql but fails validation (app !== 'Pinakes') is kept forever in the backups directory, listed by listBackups as a normal restorable 'full' backup, and is never cleaned up on failure. |
| F011 | 50 | correctness | `app/Views/admin/updates.php:1011-1023` | The PR adds admin-only 403 re-checks to downloadBackup and deleteBackup, but the backups table (whose GET route still allows staff) renders the Download and Delete buttons unconditionally — only the restore button is gated by IS_ADMIN at line 983 — so staff users see actionable buttons that now always fail with 'Operazione riservata agli amministratori'. |
| F012 | 45 | correctness | `app/Support/BackupManager.php:483-485` | dumpDatabaseTo issues SET SESSION TRANSACTION ISOLATION LEVEL REPEATABLE READ on the shared request connection ($this->db) and never restores the previous level after commit, so every query the rest of the request runs on that connection inherits REPEATABLE READ — a silent behavior change when the server default is READ COMMITTED. |
| F014 | 50 | correctness | `app/Controllers/UpdateController.php:260-271` | downloadBackup falls back to 'Content-Length: 0' when filesize() returns false while still attaching the non-empty file stream as the body, so on a stat failure the client receives a truncated (0-byte) download that looks like a successful empty backup instead of an error; omitting the header (or erroring) would be correct. |

Phase 4 couldn't confirm decisively. Re-run `/adamsreview:review` if you suspect this deserves
further investigation with fresh context.

## Fix runs

### Run `fixrun_20260610T140427Z39d6f2` — 2026-06-10T14:28:15Z

- Outcomes: 3 fixed and verified
- Commits: `f89255b`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F002 | FG-1 | ✓ fixed and verified |  |
| F003 | FG-1 | ✓ fixed and verified |  |
| F005 | FG-1 | ✓ fixed and verified |  |

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
Incluso nella release stabile **v0.7.19** (contenuto mergiato in `main`). Chiuso come parte del rilascio 0.7.19.

PR #169, opened 2026-06-12
feat(catalog): faceted filters (collapse-on-select, author facet, noise-free, theme-aware) + F006 cover fix

fabiodalez-dev:
## Public catalog — faceted filters (clear, noise-free, theme-aware) + F006 cover fix

Reworks the `/catalogo` filters into a proper faceted-search UX, and fixes a real cover bug uncovered along the way.

### Filters
- **Collapse-on-select** — once a facet is chosen (publisher, author, genre, media type) its option list collapses to a single themed pill ("Editore: Mondadori ✕") with a "Cambia" link to re-expand. No more scrolling past the other 36 publishers you've already filtered out.
- **Author facet (new)** — scoped, with per-option counts; param `autore_id`.
- **Noise removal** — facets/options that would return nothing are hidden: `tipo_media` now carries live counts and shows only present types; the year slider clamps to the real min/max of the current result set; any facet with ≤1 reachable value (and an empty genre facet) is suppressed entirely.
- **Cross-facet scoping** — every facet counts against the other active filters (extends the existing remove-self pattern to authors, media types and year).
- **Theme-aware + native look** — uses only the existing `:root` CSS variables (`--primary-color`, …), so the filters follow the active theme; styling matches the book cards / availability cards.

Backend: `FrontendController::getFilterOptions()` gains `autori`, `media_types` (with counts), `anno_bounds`, and `suppress` flags (propagated to the page and the `/api/catalogo` JSON). Frontend: `catalog.php` markup + theme-aware CSS + JS (shared facet-collapse/suppression/year-clamp helpers, active chips). No schema change.

### F006 cover fix
The F006 regression test only set `scraped_cover_url`, while the real UI (`applyAlternativeCover`, book_form.php) sets **both** `copertina_url` and `scraped_cover_url`. Making the test faithful exposed a real bug: editing a book with a **non-whitelisted** cover URL repointed `libri.copertina_url` to the dead remote URL (the old local file was kept as an orphan, but the book pointed at nothing). Now a failed external replacement keeps the existing cover.

### Tests
- New `tests/catalog-facets.spec.js` (12 read-only checks: collapse-on-select, scoped counts, suppression, year bounds, chips, clear-all) — 12/12 green against a populated catalog.
- `tests/pr166-167-coverage.spec.js` test 28 now mirrors the real UI; the full 32-test suite is green on a fresh dedicated install (incl. the F006 fix). PHPStan level 5 clean.


fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
Incluso nella release stabile **v0.7.19** (contenuto mergiato in `main`). Chiuso come parte del rilascio 0.7.19.

PR #174, opened 2026-06-15
fix(updater): fail-closed patch error handling + redirect-safe releases fetch

fabiodalez-dev:
Follow-up to the PR #172 review. Each finding verified against current code; only still-valid ones fixed.

## Fixed
- **`applyPreUpdatePatch` fail-open → fail-closed.** A present, digest-verified pre-update patch that failed to stage (`file_put_contents`), parsed to a non-array, or threw mid-apply used to return `success=true`, letting the install proceed half/un-patched. Now `success=false` (the caller aborts; nothing committed yet). Consistent with the existing `invalid` handling.
- **`applyPostInstallPatch` catch surfaces errors.** A runtime exception set only `$error` but left `success=true`, hiding it as “no patch applied”. Now `success=false`, so it surfaces through the caller’s **non-fatal warning** branch (core update is already done → stays a warning).
- **`fetchReleasesRaw` cURL branch redirect-safe.** Primary + token-retry contexts now pin `FOLLOWLOCATION=false` / `MAXREDIRS=0` and gate success to 2xx, matching `makeGitHubRequest`. (Token was already protected by `UNRESTRICTED_AUTH=false`; this is defense-in-depth/consistency. Asset-download paths keep `FOLLOWLOCATION=true` on purpose — github.com→CDN.)

## Skipped (with reason)
- Locale digest keys (it/en/de/fr), sidecar guard removal, digest full-hex regex, scrape-hooks `false→null` — **already addressed** in earlier commits.
- create-release.sh “draft-safe STEP 9 / cleanup” — **false positive**: `gh release view <tag>` (CLI) resolves drafts by tag_name (the 0.7.20 release passed those steps as a draft). Only the REST `/releases/tags` endpoint 404s on drafts, which is exactly why STEP 9.5 uses the API list.

## Validation
PHPStan L5 clean · `updater-hardening` unit **37/37** (+4 new fail-closed assertions) · all `tests/*.unit.php` green.

## Note di Rilascio

* **Bug Fixes**
  * Migliorata la robustezza del sistema di aggiornamento con gestione più rigorosa dei codici di risposta della API
  * Rafforzata la gestione degli errori durante il processo di aggiornamento per bloccare operazioni non sicure
  * Garantito il fallimento controllato delle operazioni di patching in caso di problemi

* **Tests**
  * Aggiunti test di sicurezza per verificare il comportamento del sistema di aggiornamento

### DIFF
--- a/app/Support/Updater.php
+++ b/app/Support/Updater.php
@@ -999,7 +999,11 @@ class Updater
             $ch = curl_init($url);
             curl_setopt_array($ch, [
                 CURLOPT_RETURNTRANSFER => true,
-                CURLOPT_FOLLOWLOCATION => true,
+                // Do not follow redirects on this authenticated API call (api.github.com
+                // does not legitimately redirect); a 3xx is rejected as failure below,
+                // matching makeGitHubRequest()'s fail-closed redirect stance.
+                CURLOPT_FOLLOWLOCATION => false,
+                CURLOPT_MAXREDIRS => 0,
                 CURLOPT_TIMEOUT => 30,
                 CURLOPT_CONNECTTIMEOUT => 10,
                 CURLOPT_USERAGENT => 'Pinakes-Updater/1.0',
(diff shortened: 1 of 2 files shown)


fabiodalez-dev:
@coderabbitai full review

PR #177, opened 2026-06-17
feat(mobile-api): per-user REST API + UnifiedPush for the native app

fabiodalez-dev:
## What this is

A bundled, toggle-able **Mobile API plugin** exposing everything a logged-in
library user can do on the web through a versioned REST API at `/api/v1`, so a
native app (Android first) can connect to any Pinakes instance by URL, log in
with email+password, and use search, reservations/loans, wishlist, profile,
messaging, and push notifications. Built per `MOBILE_API_SPEC.md` (the contract,
committed on this branch).

**No Android code here** — that's a separate repo. This is the backend the app consumes.

## Highlights

- **Auth**: opaque, revocable, per-device bearer tokens, **stored hashed** (sha256 + `hash_equals`), HTTPS-enforced, per-token quotas, login throttled. Device list + single-device revoke.
- **Catalog**: search (text / author / publisher / genre-cascade / language / availability), **cursor pagination**, ETag/304, full book detail + **personal history**, genre tree.
- **User actions**: loans, reservations (create + cancel, reusing core overlap/availability rules), wishlist, profile + password, contact message, in-app notifications feed.
- **Push**: UnifiedPush provider + dispatcher hooked into the existing reminder cron, per-type prefs + quiet hours, **graceful NullProvider fallback** when unconfigured (never hard-fails).
- **Discovery**: public `GET /api/v1/health` (identity, version, feature flags, app-access/registration/private-mode, https warning).
- **Docs**: OpenAPI 3.1 (`/api/v1/openapi.json`) + Swagger UI (`/api/v1/docs`), admin settings tab, i18n in all 4 locales.

## Security — SSRF model (push endpoints)

The push endpoint is user-supplied and the server POSTs to it. Hardened by reusing
`app/Support/SsrfGuard.php`:

- **Registration**: https-only, no userinfo, no bare-IP literal, port 443 only, host must resolve where **every** A/AAAA record is public (blocks loopback/RFC1918/link-local/CGNAT/ULA/NAT64/IPv4-mapped).
- **Send**: re-resolves + **pins** the connection to a vetted public IP (`HttpClient` new `pin_ip` → Guzzle `CURLOPT_RESOLVE`), `max_redirects=0` so a 302 can't bypass. Closes the TOCTOU/DNS-rebind window.

## Verification

- **PHPStan level 5**: clean (plugin + `HttpClient`) — 0 errors.
- **E2E** `tests/mobile-api.spec.js`: **72 passed, 1 skipped** on Apache :8081 + MySQL. The skip is the login rate-limit test, which self-skips while the dev server runs `PINAKES_E2E_BYPASS_RATE_LIMIT=1` (needed by the broader serial suite); it still asserts 429 on a server without the bypass.
- **Data isolation** is explicitly tested (user B can't read A's loans/wishlist/devices nor cancel A's reservation); soft-delete hiding tested.

## What to review first

1. **Security**: `PushController::isSafePushEndpoint`, `UnifiedPushProvider::send`, `HttpClient::pin_ip`.
2. **Data isolation**: every authed query scoped to the token-resolved user id.
3. **Honest caveats in `STATUS.md`**: VAPID JWT signing is advisory-only, `FcmProvider` is a stub, and the cron push *delivery* is wired but not exercised in CI. Read it before merging.

Not merging — for your review.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Introdotto il plugin **Mobile API** con endpoint **/api/v1**: discovery/health, autenticazione a token (login/register), catalogo con paginazione a cursore, prestiti/prenotazioni, wishlist, profilo, cambio password e messaggi.
  * Nuove **notifiche push** con **UnifiedPush**, gestione **sottoscrizioni** e **preferenze** (quiet hours) e dispatch schedulato con dedup/idempotenza.
  * Disponibile pannello admin per revoca dispositivi e settaggi Mobile API; enforcement **HTTPS** e **rate limit** per token.

* **Documentation**
  * Aggiunta specifica Mobile API e documentazione **OpenAPI/Swagger** live.

* **Tests**
  * Nuove suite E2E Playwright, incluse verifiche di comportamento e **idempotenza**.

* **Chores**
  * Aggiornate bundled plugin, regole di rilascio e **`.gitignore`** per i file `mobile-api` (solo sorgenti/riferimenti selezionati).
  * Aggiornati testi di localizzazione e bump di versione.


fabiodalez-dev:
## Code review — `feature/mobile-api` (38 files, 6 lenses)

Six review lenses (diff-local, structural, CLAUDE.md, comments, UX, security) over `main..HEAD`, each finding adversarially verified against the code. CLAUDE.md compliance lens: **clean** (soft-delete, escaping, admin-route literals, `\Throwable`, `ensureSchema` from both hooks, no `doAction` in `onActivate` — all respected).

### 🔴 Should fix before merge

| # | Finding | File | Why |
|---|---------|------|-----|
| 1 | **i18n gap (blocking project rule)** — ~45 API `__()` strings (auth, token, rate-limit, HTTPS, loan/reservation/wishlist errors) are **not** in any of the 4 locale JSONs | `AuthController`, `ActionsController`, `AppAuthMiddleware`, `HttpsEnforceMiddleware`, `TokenQuotaMiddleware` + `locale/*.json` | Only the push/settings strings were added. EN/FR/DE users get Italian fallback. The rule is "every `__()` string in all 4 locales, same commit". |
| 2 | **`password_confirm` guard is dead** (4 lenses) — `$p2 = $body['password_confirm'] ?? ($body['password'] ?? '')` makes the mismatch check always pass when the field is absent | `AuthController::register` (147), `ActionsController::changePassword` (580) | A client sending only `password` bypasses confirmation entirely. Either require the field or drop the check honestly. |
| 3 | **`book_available` notification can be permanently lost** | `PushDispatcher::sweepBookAvailable` (224-248) + `ActionsController` feed (686) | The watcher (the only dedup) is `DELETE`d **before** `deliver()`; `deliver()` then skips on NullProvider / quiet-hours / pref-off / no-subs, and the `/me/notifications` feed does **not** emit `book_available` (despite the docblock claiming it does). Net: user is never told. The "a skipped push still shows in the feed" guarantee holds for the other 3 events, not this one. |
| 4 | **Stateless API mutates `$_SESSION['user']`** every request | `AppAuthMiddleware` (81) | A bearer-token API call writes server-side session state; with a web cookie present it can clobber the user's concurrent web session. |

### 🟠 Worth fixing

| # | Finding | File |
|---|---------|------|
| 5 | `$this->db->affected_rows` read after a prepared statement (use `$stmt->affected_rows`) — 4 sites incl. the push idempotency `claim()` and token revoke | `PushDispatcher` (234, 460), `ActionsController` (429), `TokenService` (150) |
| 6 | `get_result()->num_rows` with no `false`-check → fatal `TypeError` on a mysqli error under `strict_types` | `AuthController` (178, 432) |
| 7 | `claim()` consumes the dedup row **before** `deliver()` checks pref/quiet-hours → push dropped for the whole quiet-hours window (feed still covers loan/reservation; compounds #3 for book_available) | `PushDispatcher` (264-273) |
| 8 | `new_message` is a stored/advertised push pref but **no code path ever sends it** (dead toggle) — known push incompleteness | `PushDispatcher::prefAllows` (312) |

### 🟡 Polish (low)

- **Docs drift** (4): docblock says "4 tables" but 5 are created; "auth/catalog/… land in later slices" is stale (all wired here); `PushDispatcher` idempotency doc describes `mobile_push_log` but `book_available` dedups via watcher `DELETE`; `AppAuthMiddleware` doc omits the `email_verificata=1` gate.
- **UX/a11y** (`views/settings.php`): flash banners need `role="alert"`/`aria-live`; per-row Revoke button needs a device-specific `aria-label`; empty actions `<th>` needs an sr-only label. `register` success message omits that admin approval is also required. `invalid_endpoint` message doesn't mention the port-443 / no-IP-literal constraints.
- **`X-Forwarded-Proto` trusted unconditionally** (`HealthController::isHttps`, `HttpsEnforceMiddleware::isSecure`) — fine behind a trusting proxy, but a no-proxy deployment lets a client assert `https`. Consider a proxy allow-list. (`HttpsEnforceMiddleware`'s **loopback** exemption was already moved to `REMOTE_ADDR` in this branch.)
- **`HttpClient` IP pin only when `curl` is loaded** — defense-in-depth gap if curl is absent (curl is a Pinakes requirement, and the registration-time `SsrfGuard` check still blocks non-public hosts, so low risk).

### ✅ Checked and cleared (not issues)

- IPv6 bracket literal `[::1]` push endpoint — **not** a bypass: `filter_var` doesn't match it but `SsrfGuard::resolvePinnedIp('[::1]')` fails to resolve, so it's still rejected.
- VAPID ES256 signing — signature verifies against the public key; DER→raw conversion correct.
- SSRF on the push send path — host re-resolved to a public IP and pinned via `CURLOPT_RESOLVE`, `max_redirects=0`.
- Data isolation — every authenticated query is scoped to the token-resolved `user_id` (covered by E2E tests 66-69).
- Soft-delete — `AND deleted_at IS NULL` on every `libri` query (L3 confirmed).

---

**Net:** the architecture and security core are sound; the must-fix list is concentrated in (a) the missing locale keys and (b) the push-dispatcher delivery/feed semantics for `book_available`. Items 1-7 are mechanical, low-risk fixes.

fabiodalez-dev:
## Review findings fixed — `02f4e0bd`

All 🔴 and 🟠 items from the review above are addressed; PHPStan L5 clean, E2E **73 passed / 1 skipped** (the rate-limit test self-skips under the server's E2E bypass).

| # | Finding | Fix |
|---|---------|-----|
| 1 | i18n gap (blocking) | All **61** missing API `__()` strings added to **all 4** locales (it source + en/fr/de). 0 untranslated strings remain. |
| 2 | `password_confirm` dead guard | `register` + `changePassword` no longer default the confirm field to the password — the mismatch check is enforced. |
| 3 | `book_available` lost when push off | Watcher cleared only **after** a delivered push (kept on NullProvider/quiet-hours/no-sub); `book_available` now emitted by `GET /me/notifications`. |
| 4 | `$_SESSION` clobbered by stateless API | `AppAuthMiddleware` mirrors identity for the request only and **restores** the prior value in a `finally`. |
| 5 | `$this->db->affected_rows` after prepared stmt (4 sites) | Switched to `$stmt->affected_rows` (incl. push `claim()` + token revoke). |
| 6 | `get_result()->num_rows` no false-check (2 sites) | Null-checked `mysqli_result` before `->num_rows`. |
| 7 | `claim()` before quiet-hours | New `shouldNotify()` gate runs **before** the dedup claim, so a quiet-hours pass re-fires later instead of dropping the push. |
| — | Docs drift ×4 + a11y | Stale docblocks corrected; `role`/`aria-live` on flash banners, per-device `aria-label` on Revoke, sr-only actions header. |

**Deferred (low, by design):** `new_message` is a wired-but-unfired trigger (no producer yet — tracked in `STATUS.md`); `X-Forwarded-Proto` trust is the conventional behind-a-proxy behavior; `HttpClient` IP-pin is curl-only (curl is a hard requirement and registration-time `SsrfGuard` still blocks). The IPv6-bracket "bypass" was a false positive (still rejected by `SsrfGuard`).

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
## Second review round (6 lenses) + fixes — `40aad24c`

Re-reviewed the post-fix branch. The lenses **verified the previous fixes as correct** (deliver() return-count, the soft-deleted+user-scoped book_available feed query, the shouldNotify-before-claim gate) and surfaced a regression my own fix introduced plus a few pre-existing issues. All fixed; PHPStan L5 clean, E2E **73 passed / 1 skipped**.

| Sev | Finding | Fix |
|-----|---------|-----|
| 🔴 | **`book_available` could double-push** under concurrent cron passes — removing the delete-first dropped the serialization the other 3 events get from `claim()` | Re-added a `claim()` dedup keyed on the watcher id, before `deliver()`; the watcher stays the durable feed record |
| 🔴 | **NullProvider `SKIPPED` was counted as a delivery failure** → `failure_count` climbed to the ≥10 exclusion threshold, disabling a user's devices on a push-unconfigured instance | Added `PushResult::isSkipped()`; `deliver()` no longer bumps failure on a skip |
| 🔴 | **`revokeOtherTokens` could revoke every device incl. the caller** if the caller token id were ever 0/absent (`id <> 0` matches all) | Abort when the caller token id is missing |
| 🟠 | Feed labelled queued `prenotato` reservations as "ready for pickup" | Aligned the feed to `da_ritirare` only (matches the dispatcher) |
| 🟠 | **FCM service-account JSON stored in plaintext** (VAPID key was encrypted) | Encrypted at rest via `SettingsEncryption`, decrypted on read |
| 🟠 | `dummyHash()` ran a full bcrypt per unknown-email login (CPU-exhaustion vector) | Process-cached a single valid dummy hash (still constant-time) |
| 🟡 | `If-None-Match: *` forced a 304 on GET; EN used British "favourites"; DE book_available used ASCII quotes; Revoke `aria-label` empty when device name unset; sub mutators not user-scoped; 3 stale docblocks | All addressed |

**Dismissed as non-issues (verified):** non-prepared sweep queries (static, param-less — no injection); `$_SESSION ?? []` (no implicit persisted session for the cookie-less app); the `$_SESSION` finally-restore (the abort-path leak requires a browser client sending both a web cookie and a Bearer token, then a fatal mid-request — the cookie-less mobile app can't trigger it; save/restore covers the normal path).

CodeRabbit was rate-limited (org credits exhausted) and produced no review this round.

PR #197, opened 2026-06-25
fix(mobile-api): self-host Swagger UI (no CDN) — /api/v1/docs was blank

fabiodalez-dev:
## Problem
`GET /api/v1/docs` (the Mobile API Swagger UI page) rendered blank: only a `.gitkeep` had shipped under `public/assets/swagger-ui/`, so the page fell back to the jsDelivr CDN — which is blocked behind an egress firewall / offline.

## Fix
Vendor Swagger UI **5.18.2** locally and **drop the CDN fallback entirely**: `assetsBaseUrl()` now always returns the local `public/assets/swagger-ui/` path. Removed the dead `SWAGGER_UI_VERSION` const and `docRoot()` method. Added a Playwright regression test (`tests/swagger-docs.spec.js`) asserting the page references local assets and never a CDN, and that the bundle actually renders the API operations.

## Follow-up fixes from the review (commits `2376fea0`, `e05e8ffb`)
- **Escaping/XSS** — `$cssUrl`/`$jsUrl` (Host-derived) are now `htmlspecialchars(ENT_QUOTES,'UTF-8')`-escaped before going into `href`/`src`, matching `$title`/`$docUrl` (the `/api/` routes are exempt from the canonical-host redirect, so the Host is attacker-influenceable).
- **Release gate** — added the two vendored assets to `create-release.sh` `CRITICAL_FILES`, so a release can't silently ship a blank docs page.
- **JS-context escaping** — `$docUrl` now uses `json_encode(JSON_HEX_TAG | JSON_UNESCAPED_SLASHES)` (correct escaper for a JS string), not `htmlspecialchars`.
- **No more blank page** — if the bundle fails to load, the inline script renders a diagnostic with a link to the raw `openapi.json` instead of an empty `#swagger-ui`.

## Verification
`php -l` + **PHPStan level 5** clean; browser check on `/api/v1/docs` → SwaggerUIBundle boots, 29 operation blocks render, 0 JS errors; `swagger-docs.spec.js` 4/4. No runtime CDN dependency remains.


fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
### Code review

**Branch:** `fix/swagger-ui-vendor-no-cdn` → `main`
**PR:** #197 (open)

Automated 6-lens review. The core of the PR — dropping the jsDelivr CDN fallback and serving Swagger UI from vendored local assets only — is correct. Below are findings on **pre-existing code** that the CDN removal makes more relevant.

#### 🔧 To fix (confirmed — escaping/XSS, rule #3)
**`SwaggerUiController.php:62-74,104`** · `$cssUrl`/`$jsUrl` (and thus `$assetsBaseUrl`, derived from the **Host** header via `baseUrl()`) are interpolated into `href="{$cssUrl}"` and `<script src="{$jsUrl}">` **without** `htmlspecialchars(ENT_QUOTES,'UTF-8')`, while `$title` and `$docUrl` in the same method ARE escaped — a real inconsistency and a CLAUDE.md rule #3 violation.
- The Host is **not** normalized on this route: the canonical-host redirect in `public/index.php` explicitly skips `/api/` paths → an arbitrary Host reaches the controller intact. A value with `"` or `>` could break out of the attribute (reflected XSS).
- **Fix:** `$cssUrl = htmlspecialchars($assetsBaseUrl.'/swagger-ui.css', ENT_QUOTES, 'UTF-8');` and likewise for `$jsUrl` (lines 62-63). Behavior-preserving for legitimate hosts.

#### ℹ️ To consider (real, below the confirmation threshold)
- **`scripts/create-release.sh:156-166`** — with the CDN fallback gone, `/api/v1/docs` now hard-depends on the two vendored `public/assets/swagger-ui/*` files, but they are **not** in `CRITICAL_FILES` (TinyMCE is). `git archive` ships them today (tracked, not export-ignored), so no immediate breakage; but a future `export-ignore`/removal would build a release with a **blank** docs page and a green build, detectable only via E2E. _Suggestion:_ add the two paths to `CRITICAL_FILES`, as already done for TinyMCE.
- **`SwaggerUiController.php:110`** — `$docUrl` is HTML-escaped (`htmlspecialchars`) but used inside a **JS string** (`url: "{$docUrl}"`): wrong escaper for the context. Low risk (the value is scheme/host/path-derived, not user input), but CLAUDE.md rule #3 prescribes `json_encode(..., JSON_HEX_TAG)` for PHP→JS.

#### No findings
- **UX** — one minor finding (blank page with no diagnostic if the bundle fails to load), classified below threshold on a developer-facing docs page.
- The 2 vendored assets (`swagger-ui-bundle.js`, `swagger-ui.css`) are upstream minified library files, out of scope for line-by-line review. `docRoot()` and `SWAGGER_UI_VERSION` removed with no other references (verified). The new E2E test covers the "self-hosted, no CDN" case.

fabiodalez-dev:
**Fixes applied** (commit `2376fea0`):

- 🔧 **Escaping/XSS** — `SwaggerUiController::buildHtml()` now wraps `$cssUrl`/`$jsUrl` (Host-derived) in `htmlspecialchars(ENT_QUOTES,'UTF-8')`, like `$title`/`$docUrl`. A malicious Host is neutralized, legitimate URLs unchanged.
- 🔧 **Release** — `create-release.sh`: added `public/assets/swagger-ui/swagger-ui-bundle.js` and `swagger-ui.css` to `CRITICAL_FILES` (already shipped via git archive; now the gate verifies them).

Verified: `php -l` clean, **PHPStan level 5** clean, `bash -n` clean, assets present + tracked (1.4 MB / 152 KB).

_Still open, not requested in this round:_ `$docUrl` in a JS context (`json_encode(JSON_HEX_TAG)` instead of `htmlspecialchars`) and the blank page with no diagnostic if the bundle fails to load — both low-risk.

fabiodalez-dev:
**Fixed the last two as well** (commit `e05e8ffb`):

- 🔧 **`$docUrl` in a JS context** — now `json_encode($openApiUrl, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES)` instead of `htmlspecialchars` (the right escaper for a JS string; `JSON_HEX_TAG` blocks a `</script>` breakout). The heredoc uses `url: {$docUrl}` (json_encode already supplies the quotes).
- 🔧 **Blank page → diagnostic** — if `SwaggerUIBundle` is undefined (bundle not loaded), the script now renders a message with a link to the raw `openapi.json` instead of leaving `#swagger-ui` empty.

Verified: `php -l` clean, **PHPStan L5** clean, and a **real-browser check on `/api/v1/docs`** → SwaggerUIBundle boots, **29 operation blocks** rendered, **0 JS errors**. Malicious Host input correctly neutralized (`\"`, `<`).

All four review findings are now closed.

fabiodalez-dev:
@coderabbitai review

PR #199, opened 2026-06-26
fix(loans): promote the waitlist head when reservations fully subscribe the copies (#157)

fabiodalez-dev:
## Problem
Reservation maintenance (`cron/full-maintenance.php` → `MaintenanceService::processScheduledReservations`) promoted **no** reservation to a pending loan when a waitlist fully subscribed a title: **1 copy + 2 active reservations → 0 promoted** instead of 1.

## Cause
The promotion gate (`ReservationManager::processBookAvailability` → `isDateRangeAvailable` → `CapacityService::hasFreeCapacity`) counted **every** overlapping active reservation as occupying a capacity unit, excluding only the one being promoted. With 1 copy and 2 reservations, promoting the head (`queue_position=1`) still counted the second (`queue_position=2`) as occupying the only copy → `occ == total` → head never promoted. Both runs converted 0; test **F.21b (#157)** saw 0 pending instead of 1.

## Fix
The promotion path now passes the promoted reservation's `queue_position`, so the capacity gate **ignores active reservations with a strictly greater `queue_position`** (lower-priority waitlist entries, promoted in later runs as copies free up).
- Reservations **ahead** in the queue (lower position) still count.
- A legacy NULL `queue_position` still counts (conservative — never overbook).
- The new parameter is optional (`default null`): the **admin-creation gate and overbooked-auditor**, which share `CapacityService`, are unchanged.

## Verification
- **PHPStan level 5** clean.
- Cron repro: run 1 → **1 pending**, run 2 → still **1** (no double-conversion).
- E2E: **loan-reservation-complete 26/26** (F.21b now passes), **loan-overlap 35/35** (overbook prevention intact), **loan-state-model 4/4**.

_Surfaced while running the full E2E suite; pre-existing on `main`, independent of the Swagger PR #197._

### DIFF
--- a/app/Controllers/ReservationManager.php
+++ b/app/Controllers/ReservationManager.php
@@ -187,7 +187,16 @@ class ReservationManager
                 // $reservation['data_fine_richiesta'] for the loan period).
                 $nextReservation['data_fine_richiesta'] = $endDate;
 
-                if ($this->isDateRangeAvailable($bookId, $startDate, $endDate, (int) $nextReservation['id'])) {
+                // #157: pass the promoted reservation's queue_position so the
+                // capacity gate ignores waitlist entries BEHIND it (they would
+                // otherwise block the head when the waitlist fully subscribes the
+                // copies — e.g. 1 copy + 2 queued reservations promoted 0).
+                // isset() is false for both an absent key and a NULL value, so it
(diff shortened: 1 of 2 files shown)


fabiodalez-dev:
@coderabbitai review

PR #239, opened 2026-07-08
test(gate): activate every activatable plugin, not just book-club

fabiodalez-dev:
The book-club activation failure (#233) reached a real install because the gate never activated the plugin. The same blind spot exists for **every other bundled plugin**, and for any added in the future.

Adds **test 6** to `plugin-integrity.spec.js`: it enumerates the plugins from the DB and activates every one that isn't active yet through the real `/admin/plugins/{id}/activate` endpoint, asserting `onActivate`/`ensureSchema` succeeds (a failure comes back as `{success:false, "Schema activation failed for: <table>"}`) and that it ends up active. The list is **read from the DB**, so a plugin added later is covered automatically.

**Why plugin-integrity and not full-test:** it's a dedicated, plugin-focused spec that already runs post-install AND post-upgrade in both CI and `scripts/reinstall-test.sh` (Test A + Test B), and it isn't hostage to full-test's 20-phase serial chain.

The three external-API plugins (discogs/deezer/musicbrainz) are excluded — **test 4** already documents that their `onActivate` throws when the CI host is offline, so they stay opt-in. (An earlier draft of this PR tried to activate *all* inactive plugins inside full-test; that would have failed on exactly those three.)

### DIFF
--- a/tests/full-test.spec.js
+++ b/tests/full-test.spec.js
@@ -136,17 +136,28 @@ async function dismissSwal(page) {
   } catch (_) { /* already closed */ }
 }
 
-/** Return today's date as YYYY-MM-DD. */
-function todayISO() {
-  const d = new Date();
-  return `${d.getFullYear()}-${String(d.getMonth() + 1).padStart(2, '0')}-${String(d.getDate()).padStart(2, '0')}`;
+// Reservation dates are validated server-side against DateHelper::today(), which
+// runs in the app timezone (Europe/Rome by default). The CI runner is UTC, so a
+// plain `new Date()` yields the UTC calendar day — and when CI runs late in the UTC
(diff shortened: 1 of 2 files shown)


fabiodalez-dev:
Bundled a second, unrelated fix into this branch because it was blocking the CI from ever going green: `full-test`'s `todayISO()`/`futureISO()` computed the reservation date in the **runner** timezone (UTC in CI) while the server validates it against `DateHelper::today()` in the **app** timezone (Europe/Rome). When CI runs late in the UTC day, UTC "today" is Rome's yesterday → the server rejects the reservation as `past_date` → Phase 15.2 "Request loan" fails. It reproduced 3/3 on 23:xx-UTC runs and passes when CI runs earlier — a pre-existing, time-of-day flake, not caused by the plugin-integrity change. Both dates are now formatted in Europe/Rome via `Intl`.

PR #241, opened 2026-07-09
fix(search): harden FULLTEXT term handling + close import/API index gaps

fabiodalez-dev:
Follow-up hardening on the denormalized FULLTEXT search (#240), found reviewing uncommitted changes. Each item fixes a real bug:

## Correctness
- **Stopwords / punctuation zeroed out results.** A required BOOLEAN-mode term `+word*` matches **nothing** when the word is a FULLTEXT stopword (`the`, `la`, `di`, `and`, …) or gets split by the parser (`C++`, `Q&A`, `L'orologio`) — those are not indexed as one token, so the AND-required term returns zero rows. Now such words are detected (approximate InnoDB tokenization + default stopword list, mb-aware length) and routed to the `LIKE` fallback. Verified live: `il signore` → "Il Signore degli Anelli".
- **LIKE-wildcard injection.** The fallback now escapes user `% _ \` with `ESCAPE '\\'`, and the upgrade-window legacy path matches the full pre-migration scope (title/subtitle/description/keywords/ISBN/EAN/author/publisher) instead of only 5 columns.

## Gaps closed
- **LibriApiController** (admin books API) still used the old OR-of-LIKE + author `EXISTS` full scan → now routed through `buildSearchCondition` so it uses the FULLTEXT index.
- **LibraryThing import** cleared a conflicting book's ISBN/EAN but left the stale identifier in that book's `search_index`. New `clearIdentifierConflicts` (column-whitelisted, soft-delete scoped) snapshots the affected ids and rebuilds their index.

## Tests
- `tests/search-index-builder.unit.php` — **9/9** (stopword `The Hobbit`/`for Crows`, `C++`, `Q&A`, Italian title).
- `tests/librarything-search-index-conflicts.unit.php` — **4/4** (conflict ISBN cleared + `search_index` rebuilt without it).
- PHPStan L5 clean; `php -l` clean.

## Summary by CodeRabbit

* **New Features**
  * Migliorata la ricerca: le condizioni di ricerca su `search_index` sono ora costruite in modo più robusto, con gestione dedicata di stopword, simboli e token FULLTEXT.
* **Bug Fixes**
  * Gestiti i conflitti tra libri durante l’import: se ISBN/EAN si sovrappongono, gli identificatori vengono ripuliti sui record in conflitto e l’indice di ricerca viene rigenerato in modo coerente.
* **Refactor**
  * Ricostruzione dell’`search_index` più efficiente tramite aggiornamenti batch.
* **Tests**
  * Aggiunti test per conflitti di identificatori e per la costruzione delle condizioni di ricerca.

### DIFF
--- a/app/Controllers/LibraryThingImportController.php
+++ b/app/Controllers/LibraryThingImportController.php
@@ -1287,24 +1287,17 @@ class LibraryThingImportController
 
         if ($existingBookId !== null) {
             $this->log("[upsertBook] UPDATING existing book ID: $existingBookId");
+            $conflictingBookIds = [];
 
             // Clear ISBNs from other books if they conflict (TSV data is authoritative)
             if (!empty($data['isbn13'])) {
-                $stmt = $db->prepare("UPDATE libri SET isbn13 = NULL WHERE isbn13 = ? AND id != ? AND deleted_at IS NULL");
-                $stmt->bind_param('si', $data['isbn13'], $existingBookId);
-                $stmt->execute();
-                $conflictsCleared = $stmt->affected_rows;
-                $stmt->close();
+                $conflictsCleared = $this->clearIdentifierConflicts($db, 'isbn13', (string) $data['isbn13'], $existingBookId, $conflictingBookIds);
                 if ($conflictsCleared > 0) {
                     $this->log("[upsertBook] Cleared ISBN13 '{$data['isbn13']}' from $conflictsCleared conflicting book(s)");
                 }
             }
             if (!empty($data['isbn10'])) {
-                $stmt = $db->prepare("UPDATE libri SET isbn10 = NULL WHERE isbn10 = ? AND id != ? AND deleted_at IS NULL");
-                $stmt->bind_param('si', $data['isbn10'], $existingBookId);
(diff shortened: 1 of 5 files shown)


fabiodalez-dev:
Addressed in 0370e354:

- **rebuildMany O(N)** — now one chunked `UPDATE ... WHERE l.id IN (...)` per 500 books instead of one per book, with `group_concat_max_len` set once; `rebuild()` is a thin wrapper over `rebuildMany([$id])` so single-book saves produce the identical value through one SQL path. Verified: search-index-builder 9/9, librarything-conflicts 4/4 (exercises the batch), migration 24/24.
- **ltSicCleanup FK-safety** — deletes `libri_collane`/`libri_editori`/`libri_autori` child rows before `libri`.

Skipped, with reason:
- **LIKE escaping** — already handled: `buildLegacyCondition` and the short-token fallback both use `likePattern()` + `ESCAPE '\'` (the suggestion quotes the pre-hardening base, not the current code).
- **Test-helper duplication** — left as-is; the two unit tests are standalone by design and a shared util for two files is marginal.

Issue #255, opened 2026-07-16
Configurable registration fields


Himura2la:
Currently the Phone and Address field are hardcoded as required:

https://github.com/fabiodalez-dev/Pinakes/blob/63f5c0481068f68ec57f3a0a59744e2b44286316/app/Controllers/RegistrationController.php#L67-L70

Which is very inconvenient for small communities where we can't force members to disclose their personal information. Please, make a separate database table to allow NULLS in any field. 

Also NULLS should be allowed in a `cognome` field, since we don't want to force users use their real names. This would require schema migration:

https://github.com/fabiodalez-dev/Pinakes/blob/63f5c0481068f68ec57f3a0a59744e2b44286316/installer/database/schema.sql#L973

After it's done, the second step is to allow adding custom fields. For example, we need a Telegram username, because that's the main platform for our community, and we're going to use it in API for notifications, etc.

fabiodalez-dev:
Good timing — I'm picking this up now. Plan, in three parts, all landing in one PR on a dedicated branch:

1. **Configurable required fields.** \`telefono\` and \`indirizzo\` are already nullable in the schema — only the registration controller hardcodes them as required. I'm adding per-field toggles under Settings → Registration (defaults keep today's behavior, so existing installs don't change until the admin opts out).

2. **Optional surname.** \`utenti.cognome\` becomes nullable via a guarded, idempotent migration, with the registration/profile/admin forms and the display paths (loan emails, member lists, exports) updated to render single-name users cleanly.

3. **Custom registration fields.** Admin-defined fields (label, type, required, visibility) stored in two new tables — field definitions plus per-user values — rendered in the registration form and the user profile, editable by the admin, and exposed through the API user payload so integrations like your Telegram-based notifications can read them.

It will target the release after 0.7.36 (the migration is version-gated accordingly). I'll link the PR here once it's up.

fabiodalez-dev:
PR is up: #256. All three parts are in — per-field requirement toggles, optional surname, and admin-defined custom fields (your Telegram case is the E2E test's exact scenario). One deviation from the original plan, explained in the PR: `cognome` stays NOT NULL and an optional surname is stored as an empty string, because 70+ display paths CONCAT nome+cognome and a NULL would blank the whole name — the functional result (surname not required anywhere) is the same.

Himura2la:
wow you really did all that? I'm totally amazed!

fabiodalez-dev:
[...]

PR #265, opened 2026-07-17
Fix custom CSS injection on auth frontend pages

### LINKED ISSUE #262: Custom CSS does not apply on the /register page
I want to hide some fields from the registration form using the following CSS
```css
form[action="/register"] > div:has(#indirizzo),
form[action="/register"] > div:has(#telefono),
form[action="/register"] > div > div:has(#data_nascita),
form[action="/register"] > div:has(#cod_fiscale) {
  display: none
}
```

but the "Custom CSS" feature in `/admin/settings?tab=advanced#advanced` does not insert the specified CSS onto the /register page

TBH, I don't see the "Custom CSS" feature working at all

Himura2la:
Fixes #262 

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Aggiunto il supporto per applicare CSS personalizzato alle pagine di accesso, registrazione, recupero e reimpostazione della password.
  * Il CSS configurato viene visualizzato solo quando presente e viene normalizzato prima dell’applicazione.
  * Estesa l’applicazione degli stili personalizzati anche al layout frontend.

### DIFF
--- a/app/Views/auth/forgot-password.php
+++ b/app/Views/auth/forgot-password.php
@@ -19,6 +19,7 @@ $forgotPasswordRoute = route_path('forgot_password');
     <style>
         body { font-family: system-ui, -apple-system, sans-serif; }
     </style>
+    <?php require __DIR__ . '/partials/custom-css.php'; ?>
 </head>
 <body class="bg-gray-50 dark:bg-gray-900">
 
--- a/app/Views/auth/login.php
+++ b/app/Views/auth/login.php
@@ -25,6 +25,7 @@ $forgotPasswordRoute = route_path('forgot_password');
     <style>
         body { font-family: system-ui, -apple-system, sans-serif; }
     </style>
+    <?php require __DIR__ . '/partials/custom-css.php'; ?>
 </head>
 <body class="bg-gray-50">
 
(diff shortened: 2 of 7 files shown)


fabiodalez-dev:
I've reimplemented this in #273 and am closing this PR in its favour — the feature itself (custom CSS reaching the auth pages) is a good call and #273 keeps it, with credit to you.

The reason I didn't merge this branch directly: it emits the CSS through `ContentSanitizer::normalizeExternalAssets()`, which only normalizes external font URLs. It does not strip a `</style><script>…</script>` sequence, so a stored `advanced.custom_header_css` value could break out of the `<style>` context and run JavaScript. An earlier fix moved the app to `ContentSanitizer::sanitizeCustomCss()` (which strips `<style>`/`<script>` and HTML comment markers) at both save and render; this branch is based on the pre-fix tree, so merging it would have undone that — and exposed it on the unauthenticated login/register pages.

#273 uses `sanitizeCustomCss()` in a shared partial, includes it on all five auth views, and refactors `frontend/layout.php` onto the same partial (no duplicate block). I verified with an injected `</style><script>` payload that it renders as inert CSS text while legitimate CSS still applies on `/register`.

If you want to hide the #255 registration fields, once #273 ships your CSS in Settings → Advanced → Custom CSS will apply on `/register`.

Issue #237, opened 2026-07-08
Managing Authors/Illustrator


ctariel:
Hello,

For some comic books, I tried separating the illustrator from the author. However, I don't have the same input assistance for the illustrator. Furthermore, the illustrator does not appear in the list of authors. 

I don't know how improve this system, and I'm not even sure it is worth. Sometime, I even have a colorist...

ctariel:
Hello, I'm still on authors. The form asks first for the real name, then the pseudonym. BUt then there is no way to find auhor with the pseudonym : 

<img width="858" height="408" alt="Image" src="https://github.com/user-attachments/assets/c6a455a9-fdd3-4d13-8f91-a072b92866b8" />
On this exemple, we can't just find "Leo".

Next, the book's entry displays the real name, whereas one would expect to find the pseudonym instead :

<img width="618" height="268" alt="Image" src="https://github.com/user-attachments/assets/3c10a899-8afd-49f7-bca9-aa4e992aede2" />

<img width="353" height="534" alt="Image" src="https://github.com/user-attachments/assets/e6d01206-bf24-41b3-8767-34307088f4cb" />

fabiodalez-dev:
Shipped in **v0.7.37** (stable, out now). Illustrators, translators, curators and the new **colorist** are proper author entities with the same search-as-you-type picker as the main author — reusable, findable by pseudonym, and shown on the contributor's author page and on the books. Existing free-text values are converted to entities automatically on the first run after upgrading, so there's nothing to redo. Books also display the pseudonym as *"Pseudonym (Real name)"* instead of only the real name. Update from **Settings → Updates**, or download the ZIP from the releases page.

HansUwe52:
This is a very useful improvement of Pinakes, But when you look at the form:
<img width="1021" height="277" alt="Image" src="https://github.com/user-attachments/assets/c93cefed-d9c6-4e75-a56a-b970b65a2c2d" />
the help text below the fields: Illustrator, Translator, Curator and Colorist should be adjustet to:

**Illustrator:** Search for an existing illustrator or add a new one by typing the name
**Translator** Search for an existing translator or add a new one by typing the name
**Curator** Search for an existing curator or add a new one by typing the name
**Colorist** Search for an existing colorist or add a new one by typing the name

HansUwe52:
the help text below the fields: Illustrator, Translator, Curator and Colorist shows now for all fields:
**Search for an existing author or type a new one**
while it shoud be
[...]

PR #273, opened 2026-07-20
Apply custom CSS to auth pages (safely) + role-specific contributor help

### LINKED ISSUE #262: Custom CSS does not apply on the /register page

fabiodalez-dev:
Closes #262. Addresses #237. Supersedes #265.

## Custom CSS on auth pages (#262 / #265)

`advanced.custom_header_css` only applied to the app chrome, so it never reached the login/register/forgot/reset pages — a user who wants to hide the #255 registration fields via CSS couldn't (#262). This adds a shared partial `app/Views/auth/partials/custom-css.php`, includes it on all five auth views, and refactors `frontend/layout.php` to use it (removing its inline duplicate).

This reimplements the intent of #265 (by @Himura2la) **safely**. #265 emitted the CSS through `ContentSanitizer::normalizeExternalAssets()` (fonts-only), which does **not** strip a `</style><script>…` breakout — merging it would have reintroduced the stored-XSS that was fixed earlier, and this time on **unauthenticated** pages. The partial uses `ContentSanitizer::sanitizeCustomCss()` (the same render-time sanitizer `frontend/layout.php` already uses), which strips `<style>`/`<script>` and HTML comment markers.

Verified live: an injected `</style><script>` payload renders as inert CSS text (window flag `undefined`, page title unchanged, 0 script tags in `<head>`), while legitimate CSS still applies on `/register` and `/accedi`. The existing `custom-css-injection.spec.js` still passes.

## Role-specific contributor help (#237, @HansUwe52)

Illustrator/Translator/Curator/Colorist showed the same generic "search for an existing author" hint. Each now has its own role-named help string, translated across all four locales (it/en/de/fr).

## Tests

`tests/auth-custom-css-and-contributor-help.unit.php` (17 assertions) locks in the safe sanitizer, the partial inclusion on every view, and the role-help keys across all four locales.

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Aggiunto il supporto a CSS personalizzati nelle pagine di accesso, registrazione e recupero password, con applicazione coerente anche nella layout principale.

* **Miglioramenti**
  * Testi di aiuto aggiornati e resi specifici per ruolo per illustratori, traduttori, curatori e coloristi.
  * Aggiunte nuove traduzioni (it-IT, en-US, de-DE, fr-FR) per i suggerimenti e i prompt di ricerca/creazione dei ruoli.


Himura2la:
btw, what's the real issue with the `</style><script>`? Next to Custom CSS feature is Custom JS, and the write access to both is only granted to instance admins, so plain users should have no ability to inject anything to other users under normal conditions with these fields....

So how this stored-XSS can even be exploited? And what's the difference of  `</style><script>...` and a legitimate "Essential JavaScript"?

Or am I miss something?

fabiodalez-dev:
Fair point, and you're mostly right. Under the current model this isn't a privilege-escalation XSS: `advanced.custom_header_css` is admin-only write (CSRF-gated), and the Custom JS field right next to it runs arbitrary JS by design. An admin who breaks out of `<style>` with `</style><script>` gains nothing they couldn't already do through the JS field, and a plain user can't write either field — so there's no lower-privilege attacker. "Stored XSS" was stronger wording than it deserves in the classic sense; it's really field-contract hardening.

Where it matters concretely, and specifically because of this PR:

Before #273, both `custom_header_css` and Custom JS lived only in `frontend/layout.php`. This PR now also renders the CSS on the **unauthenticated** auth pages (login/register/forgot/reset). The Custom JS feature does **not** run there — it's emitted only in `frontend/layout.php`, cookie-consent gated (0 occurrences in the five auth templates). So on the pre-login pages there is deliberately no JS-execution path at all.

If the partial emitted the CSS unsanitized, a `</style><script>…` value would be the one way to get script running on those pre-auth pages — for every anonymous visitor to the login screen — a surface the JS field never reaches. That's an actual expansion of where script can run, introduced by this PR. So the partial runs it through `sanitizeCustomCss()` (the same sanitizer `frontend/layout.php` already uses) to keep the "CSS" box doing only CSS there.

Difference from "Essential JavaScript": that field is an explicit, labeled, cookie-category-aware opt-in to run JS, scoped to the frontend layout. The CSS field is labeled CSS and now renders on more surfaces including pre-auth; sanitizing keeps enabling this feature from silently turning the CSS box into a JS channel on pages the JS box was never wired into.

So: agreed it's not cross-user XSS — but for a field that now renders pre-auth I'd rather it can only ever style. Shout if you see a hole in that.

Himura2la:
@fabiodalez-dev thanks for the detailed explanation, but what I don't understand yet is "why is it a problem to allow custom js on unauthenticated pages?" If an instance admin uses the XSS for injecting custom js into its pages bypassing cookie-category-aware consent, then it's purely their own responsibility for breaking GDPR on his instance via a dirty hack.

They also can modify source code to get the same result, it would be just a little bit harder...

Am I still missing something? Should you really take responsibility for admin's misuse of your interface?

fabiodalez-dev:
You're not missing anything — you're right. For a trusted admin (the only one who can write either field, who can already run JS via the Custom JS box or by editing source), this isn't a privilege boundary being crossed. There's no untrusted actor and no victim whose trust is violated: an admin serving their own script on their own pages is exactly what the Custom JS feature does by design. So it isn't a security vulnerability, and calling it "XSS" earlier was too strong — my bad.

What actually justifies `sanitizeCustomCss()` in the partial is narrower: consistency. `frontend/layout.php` already runs custom CSS through that same sanitizer (that shipped in 0.7.38), so the partial matches the existing render path instead of diverging. If the partial used the fonts-only normalizer, custom CSS would behave differently on the auth pages than on the rest of the public site — `</style>` stripped in one place, not the other. Keeping it identical is cheap and lossless for any legitimate CSS.

Whether the app should sanitize custom CSS *at all* — "it's the admin's instance, let them shoot their own foot" — is a fair position, but that's a question about the existing frontend behavior, not this PR. #273 just keeps the two paths consistent. If you think the sanitization should be dropped as theater against a non-threat, open an issue and I'll weigh it on its own merits — I'd rather change both paths together than have them disagree.

So: agreed, it's not my job to police an admin's self-misuse, and this isn't guarding against that. It's only keeping one render path from behaving differently than the one already shipped.

PR #277, opened 2026-07-21
Mobile API: public GET /auth/registration-fields for configurable/custom signup fields (#255)

fabiodalez-dev:
Adds the missing **discovery endpoint** so the Android app (and any client) can render the registration form for #255's configurable + custom fields.

### Why
`POST /auth/register` already validates custom fields and `GET /me` returns them for logged-in users, but there was **no public endpoint** exposing the field schema *before* signup — so a client couldn't know which built-in fields are required on this instance or which admin-defined custom fields to show.

### Endpoint
`GET /api/v1/auth/registration-fields` — public (no token), gated on the same app-access flag as the other auth endpoints. Returns:
- `registration_enabled` — lets the client distinguish "app disabled" from "signup closed"
- `builtin_fields` — the three config-driven toggles (`cognome`/`telefono`/`indirizzo`), each `{ required: isRequired(key), configurable: true }`
- `custom_fields` — `RegistrationFields::apiDefinitions()`: active fields only, `{ id, label, type, required }`, ordered, leaking no `attivo`/`ordine`

`data_nascita`/`cod_fiscale`/`sesso` are intentionally **not** in the signup contract (register() never reads them; profile-only, via `GET`/`PATCH /me`).

### Implementation
Reuses existing `RegistrationFields::{BUILTIN_TOGGLES, isRequired, apiDefinitions}` — no new controller file. OpenAPI spec + `mobile-api-idempotency` manifest test updated; mobile-api plugin `1.3.0 → 1.4.0`.

### Verified (live, on the dev install)
- `GET` returns **200** with the correct `{data, meta, error}` envelope.
- `builtin_fields` reflects the real `isRequired()` config.
- `custom_fields` reflects `apiDefinitions()` dynamically — seeded a field → it appeared, deleted it → gone.
- `php -l` + **PHPStan level 5** clean; route ↔ OpenAPI ↔ manifest consistent.

The matching **Android client** work (models, `PinakesApi`, dynamic register/profile fields) is in a separate PR on the `Pinakes-Android` repo.

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Aggiunto un endpoint pubblico per scoprire i campi disponibili durante la registrazione.
  * Ora sono indicati campi integrati, campi personalizzati, obbligatorietà e stato della registrazione.
  * L’endpoint è documentato nell’API ed è soggetto a limitazione delle richieste.

* **Correzioni**
  * Gestiti gli errori di accesso, le richieste eccessive e gli errori interni con risposte appropriate.

* **Manutenzione**
  * Aggiornata la versione del componente Mobile API alla 1.4.0.

### DIFF
--- a/storage/plugins/mobile-api/MobileApiPlugin.php
+++ b/storage/plugins/mobile-api/MobileApiPlugin.php
@@ -428,6 +428,17 @@ class MobileApiPlugin
                 return (new AuthController($db))->register($request, $response);
             })->add(new RateLimitMiddleware(5, 3600, 'mobile_register'));
 
+            // Public registration discovery — no token. Advertises whether
+            // self-registration is enabled and which fields (config-driven
+            // built-ins + admin-defined custom) the signup form should render.
+            // Light rate-limit to keep it from being scraped in a loop.
+            $group->get('/auth/registration-fields', function (
+                ServerRequestInterface $request,
+                ResponseInterface $response
+            ) use ($db): ResponseInterface {
+                return (new AuthController($db))->registrationFields($request, $response);
+            })->add(new RateLimitMiddleware(30, 300, 'mobile_registration_fields'));
+
             $group->post('/auth/forgot-password', function (
                 ServerRequestInterface $request,
                 ResponseInterface $response
--- a/storage/plugins/mobile-api/plugin.json
+++ b/storage/plugins/mobile-api/plugin.json
@@ -2,7 +2,7 @@
   "name": "mobile-api",
   "display_name": "Mobile API",
   "description": "API REST/JSON versionata (/api/v1) per l'app companion mobile di Pinakes: discovery, autenticazione a token per dispositivo, ricerca catalogo, prestiti/prenotazioni, wishlist, profilo, messaggi e notifiche push. Disattivata per default finché non viene abilitata.",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "author": "Fabiodalez",
   "author_url": "",
   "plugin_url": "",
--- a/storage/plugins/mobile-api/src/Controllers/AuthController.php
+++ b/storage/plugins/mobile-api/src/Controllers/AuthController.php
@@ -322,13 +322,23 @@ final class AuthController
      * Public registration discovery (issue #255).
      *
      * Advertises whether self-registration is enabled and which fields the
-     * signup form should render: the config-driven built-ins plus the
-     * admin-defined custom fields. No token — it only reveals field
-     * labels/required flags the registration form shows anyway and leaks no
-     * user data. Gated on the same app-access flag as the other public auth
-     * endpoints; the registration toggle is surfaced as `registration_enabled`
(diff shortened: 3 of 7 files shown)


fabiodalez-dev:
## 🔎 Adams Review — PR #277 (mobile registration-fields)

Adversarial multi-lens pass (correctness / structural blast-radius / security). Endpoint is correct, safe, and well-scoped — but the structural lens surfaced a real redundancy worth deciding on before merge.

### ⚠️ Confirmed — duplicate public source of truth for the registration contract
The public `GET /api/v1/health` endpoint **already** returns the same registration-discovery data this PR adds, just in a **divergent shape**:

| | `/health` (existing) | `/auth/registration-fields` (this PR) |
|---|---|---|
| enabled flag | `registration_enabled` | `registration_enabled` |
| built-in required | `registration.require_cognome` / `require_telefono` / `require_indirizzo` (flat booleans) | `builtin_fields.<field>.{required, configurable}` (nested) |
| custom fields | `registration.custom_fields` = `apiDefinitions()` | `custom_fields` = `apiDefinitions()` |

So a client now has **two** public ways to learn the same signup contract, in two shapes it must reconcile. `HealthController.php:93-97`.

**Decide one of:**
- **Drop the new endpoint**, have the Android register screen read the `registration` block from `/health` (it already calls `/health` on startup) — smallest surface.
- **Keep it** (a dedicated, purpose-named endpoint is arguably cleaner for the register screen) but consider deprecating the `registration` block in `/health`, or at least documenting that the two are intentional and which is canonical.

Not a bug — but two public contracts for one thing tends to drift. Your call.

### ℹ️ Minor — discovery payload omits the always-required core fields
`register()` unconditionally requires `nome`/`email`/`password`, but the discovery payload advertises `cognome` (in `builtin_fields`) and the custom fields, **not** `nome`/`email`/`password`. A client rendering the form purely from this contract could omit `nome`. In practice every client hardcodes name/email/password, so this is low — but a one-line note in the payload or docs ("core fields nome/email/password are always required and not listed here") would close the gap. `AuthController.php:355`.

### ✅ Clean
- **Security (L6):** public + no-token is intentional and safe — schema-only disclosure equivalent to the public `/registrati` web page; no PII/user values, `apiDefinitions()` correctly hides `attivo`/`ordine`, same app-access 403 gate as the other public auth endpoints, rate-limited 30/300s.
- **Correctness (L1):** the one flagged item — manifest row `kind:'doc'` asserting HTTP 200 while the handler can 403 — is a **false positive**: the test harness seeds `mobile_api.enabled=1` (`mobile-api-idempotency.spec.js:86`), so the endpoint always returns 200 under test.
- Route correctly public inside `/api/v1`, envelope matches `ResponseEnvelope::success`, required flags can't drift from `register()` (both call `isRequired()`), OpenAPI ↔ route ↔ manifest consistent.

_Lenses: L1 diff-local, L2 structural, L6 security. Findings verified against the code, not just flagged._

PR #278, opened 2026-07-21
Related-books section: responsive columns, no more enormous covers on large screens

fabiodalez-dev:
The **"Potrebbero interessarti"** section on the frontend book page looked terrible on large screens — the book covers were enormous.

### Cause
It used Bootstrap `col-lg-4 col-md-6` with **no cap beyond `lg`**, inside the theme's ultra-wide `.container` (~2355px on a 2560px monitor). With at most 3 related books that meant 3 columns of ~440px+, and since each cover is `padding-top: 140%` of the card width, the covers ballooned to ~615px tall on large displays.

### Fix
- **Cap the card** (`.related-book-card`) to a book-sane **280px** (matching the catalog card size), centred in its column — the cover can no longer balloon.
- **Responsive column count by screen size**: `col-12 col-sm-6 col-lg-4` → **1 / 2 / 3 columns** as the screen grows, with `justify-content-center`.
- **Group the cards**: constrain the row (`.related-books-row`, max-width 960px, centred) so the three cards sit together in the middle instead of spreading across the wide container.

### Verified live
At 375 / 768 / 1024 / 1440 / 1920 / 2560px: cards stay book-sized (280px on desktop, ≤400px on mobile via the existing mobile rule), columns scale 1→2→3, covers no longer enormous. PHPStan clean.

_Note: the section shows up to 3 related books (`getRelatedBooks` limit). If you'd like more columns to be meaningful on very wide screens, that would mean raising that limit — happy to do it if you want._

### DIFF
--- a/app/Views/frontend/book-detail.php
+++ b/app/Views/frontend/book-detail.php
@@ -1473,6 +1473,16 @@ $additional_css = "
     }
 
     /* Related Books Section */
+    /* Keep the (max 3) related cards grouped and centred instead of spread
+       across the theme's ultra-wide container. ~3 cards of 280px + gutters.
+       !important overrides the theme's wide .container/.row max-width for
+       this one scoped row only. */
+    .related-books-row {
+        max-width: 960px !important;
+        margin-left: auto !important;
+        margin-right: auto !important;
+    }
+
     .related-book-card {
         background: var(--white);
         border-radius: 16px;
@@ -1482,6 +1492,13 @@ $additional_css = "
         height: 100%;
         display: flex;
         flex-direction: column;
+        /* Cap the card to a book-sane width and centre it in its column so the
+           cover (which is 140% of the card width) never balloons on wide
+           screens where each column is very wide. Matches the catalog card
+           size (~280px). */
+        max-width: 280px;
+        margin-left: auto;
+        margin-right: auto;
     }
 
     .related-book-card:hover {
@@ -2329,9 +2346,9 @@ ob_start();
             <i class="fas fa-lightbulb"></i>
             <?= __("Potrebbero interessarti") ?>
         </h2>
-        <div class="row g-4">
+        <div class="row g-4 justify-content-center related-books-row">
             <?php foreach($related_books as $related): ?>
-            <div class="col-lg-4 col-md-6">
+            <div class="col-12 col-sm-6 col-lg-4">
                 <div class="related-book-card">
                     <div class="related-book-image-container">
                         <?php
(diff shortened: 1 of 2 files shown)


fabiodalez-dev:
## 🔎 Adams Review — PR #278 (related-books responsive)

Adversarial pass (structural + UX). The fix does solve the reported problem — covers no longer balloon on large screens (verified 375→2560px). Three **minor, non-blocking** polish items surfaced; none regress the fix.

### ℹ️ UX — size inversion at the 768px breakpoint (confirmed by measurement)
The pre-existing `@media (max-width:768px)` rule caps the card at **400px**, but the new base rule caps it at **280px**. So the card is allowed to be *wider on mobile than on desktop*: measured 760px → **310px** card, 780px → **280px** card. Widening the window just past 768px visibly **shrinks** the card ~30px instead of growing it. Small, but it reads as a glitch to a sharp eye. Aligning the two caps (e.g. drop the 400px mobile rule, or scale both from one scheme) removes it. `book-detail.php:1499` vs `:1622-1627`.

### ℹ️ UX — "matches the catalog card size (~280px)" is imprecise
The catalog grid uses `repeat(auto-fill, minmax(280px, 1fr))` — there **280px is a floor that stretches** to fill the track, so real catalog cards are usually wider than 280px. This PR's hard `max-width:280px` makes the related cards read **narrower** than catalog cards on the same site. Cosmetic (different pages, rarely side-by-side), but the code comment overstates the parity. `book-detail.php:1499`.

### ℹ️ Structural — `!important` margins fight Bootstrap's row gutters
`.related-books-row { margin-left/right:auto !important }` sits on the same element as Bootstrap's `.row`, overriding its negative gutter margins (`calc(-.5 * var(--bs-gutter-x))`), so the outer columns gain a ~0.75rem inset within the 960px cap. The measured result still looks right (3× 280px cards, 40px gaps, centred), so it's harmless — but the clean approach is a **wrapper div** carrying the max-width/centering, leaving `.row` untouched. `book-detail.php:1480-1484`.

### ✅ Clean
- Bundled CSS is Bootstrap 5.3.8; `col-12`/`col-sm-6`/`col-lg-4`/`justify-content-center` all exist. Base 280px cap and the mobile 400px rule compose without a specificity fight (media query wins ≤768px). `.related-books-row` is used only here, on the correct element; section guarded by `if (!empty($related_books))`.
- Responsive column count 1→2→3 verified live at 375/768/1024/1440/1920/2560px.

**Verdict:** ship-ready — the covers-are-enormous bug is fixed. The three items above are optional refinements; the 768px inversion is the one most worth a follow-up. Want me to apply the wrapper-div + single-cap cleanup?

_Lenses: L2 structural, L5 UX. Findings verified against live rendering, not just flagged._

Issue #286, opened 2026-07-23
genere_id/sottogenere_id vs. genere_id_cascade/sottogenere_id_cascade


Himura2la:
https://github.com/search?q=repo%3Afabiodalez-dev%2FPinakes+%22id_cascade%22&type=code

why these fields exist? seems like leftovers

fabiodalez-dev:
These are not leftover columns — the `libri` table only ever has `genere_id` and `sottogenere_id` (see `installer/database/schema.sql`). There is no `genere_id_cascade` / `sottogenere_id_cascade` column in the schema or in any migration, and nothing is ever written to or read from such a column.

`*_cascade` are transient, in-memory array keys computed at read time by `BookRepository::resolveGenreHierarchy()`. A book stores its genre as a single `genere_id`, which can legitimately point at any of the three cascade levels (root → genre → subgenre). To pre-populate the three cascading dropdowns on the edit form, that method walks the `generi` self-referential parent tree upward and maps the stored ID onto the correct level, exposing it as `genere_id_cascade` (L2) / `sottogenere_id_cascade` (L3). `book_form.php` then prefers the resolved cascade value and falls back to the raw column:

```php
'genere_id'     => (int)(($book['genere_id_cascade']     ?? null) ?: ($book['genere_id']     ?? 0)),
'sottogenere_id'=> (int)(($book['sottogenere_id_cascade']?? null) ?: ($book['sottogenere_id'] ?? 0)),
```

Without these the edit dropdowns would fail to pre-select the right level for a book whose `genere_id` isn't stored at the canonical middle level.

PR #287 additionally uses the same resolved cascade IDs to fix the genre breadcrumb on the admin book-detail page (a root+child genre was rendering duplicate/empty links). So the fields are load-bearing, not cruft. Closing this once #287 is merged.

Himura2la:
> A book stores its genre as a single genere_id, which can legitimately point at any of the three cascade levels (root → genre → subgenre).

This explains much, thank you!

Himura2la:
But still won't explain root_id and sottogenere_id.
If a book stores only genre_id as a single genre indicator, and the full chain is restored in *_cascade keys, than what's the purpose of root_id and sottogenere_id in book's metadata?

Himura2la:
Overall this part seems tangled and misleading, which brings to bugs like #287 (the fix I still don't fully understand why we can simply assume that the _cascade fields are simply a higher priority versions of plain fields).
[...]

PR #290, opened 2026-07-24
Rename computed genre keys to resolved_* (+ escape breadcrumb label)

fabiodalez-dev:
Follow-up to discussion #286 and PR #287.

The book genre hierarchy exposed two **request-only computed** keys —
`genere_id_cascade` / `sottogenere_id_cascade` — right next to the **stored**
`libri` columns `genere_id` / `sottogenere_id`, with nothing to signal that the
former are a derived view rather than persisted duplicates. That ambiguity is
what read as "tangled and misleading" in #286.

### Change
- Rename `genere_id_cascade` → **`resolved_genere_id`**, `sottogenere_id_cascade`
  → **`resolved_sottogenere_id`**. The `resolved_` prefix marks them as a
  computed view of the stored ids.
- `resolveGenreHierarchy()` (producer) + both consumers — the edit-form cascade
  pre-population and the admin book-detail breadcrumb (from #287) — updated
  together. The keys are purely in-memory (never a column, never serialized to
  an API), so the change is contained to three files.
- Documented the convention on `resolveGenreHierarchy()`.
- Switched the breadcrumb label to `htmlspecialchars(..., ENT_QUOTES, 'UTF-8')`
  (the view standard), resolving the remaining CodeRabbit nit on #287.

### Not renamed (on purpose)
`radice_id` / `radice_nome` / `genere_nome` / `sottogenere_nome` are also base
`getById()` SELECT aliases consumed by `LibriApiController`, `PublicApiController`,
the UNIMARC/MARC exports and the LibraryThing import — renaming them would break
those contracts. Only the two purely-computed `_cascade` keys are in scope.

### Verification
Live check on a book stored at root+direct-child (`genere_id=root`,
`sottogenere_id=child`): the breadcrumb renders `Root → Child` with the correct
distinct ids, not the old duplicate/empty chain. php -l + PHPStan level 5 clean;
the Phase 18 genre regression in `full-test.spec.js` covers it.


fabiodalez-dev:
### Code review

**Branch:** `refactor/genre-resolved-keys` → `main`
**PR:** #290 (open)
**Review ID:** `rev_20260724T112856Zefcc5b`
**Base freshness:** local `main` was behind `origin/main` at run start; fast-forwarded before review
**Fix threshold:** not yet set (run `/adamsreview:fix [threshold]` to apply fixes)
**Sub-agent tokens:** 442,449 across 4 invocations

Found 1 finding across all lanes:

---

---

Issue #292, opened 2026-07-25
Unable to upload Hero Background Image


Himura2la:
> Content-Security-Policy: The page’s settings blocked the loading of a resource (connect-src) at blob:https://books.hackem.cc/bf8e38d6-5b40-40c7-b050-659c5598a5f1 because it violates the following directive: “connect-src 'self'”

> Error converting file: TypeError: NetworkError when attempting to fetch resource.

at
```js
// Create a File object from Uppy file data
fetch(file.data instanceof File ? URL.createObjectURL(file.data) : file.preview)
```

fabiodalez-dev:
Fixed in v0.7.42. The hero-background uploader did `fetch(URL.createObjectURL(file.data))`, and a strict `connect-src` policy blocks `blob:`, so the converted file never reached the form. Uppy already hands over a `File` in `file.data`, so it now uses that directly — no `blob:` fetch — and the upload works regardless of the site's CSP. Verified with an automated upload test. Update from **Admin → Updates**.

Himura2la:
Now it's complete silence when I try to upload an image. No requests or console activity after choosing an image.
And it's saved with a broken link:
<img width="1909" height="940" alt="Image" src="https://github.com/user-attachments/assets/a636826b-5691-4d93-a7ce-1e305aacb5be" />

Firefox 154

Himura2la:
Same on Edge 150. The image:
<img width="409" height="307" alt="Image" src="https://github.com/user-attachments/assets/c5f84f9b-26da-45a2-92ea-66285f7c0224" />

Himura2la:
It just walks through this branch, and nothing happens (no preview, no actual upload)

https://github.com/fabiodalez-dev/Pinakes/blob/960434ec46ea7ed16281fb38039a393653d2c393/app/Views/cms/edit-home.php#L739-L741

Himura2la:
Form POST request looks like this:
```
------geckoformboundarye1ef8a389133e57847d4eff8b67b1385
Content-Disposition: form-data; name="csrf_token"

c6aa0a126137348de44a538ec88e81f8c3425731a9a09172b42f6442c1836bdc
------geckoformboundarye1ef8a389133e57847d4eff8b67b1385
Content-Disposition: form-data; name="hero[title]"

HackEm Books
------geckoformboundarye1ef8a389133e57847d4eff8b67b1385
Content-Disposition: form-data; name="hero[subtitle]"

a public library in Hacker Embassy
[...]

PR #295, opened 2026-07-26
Remove dormant fetch(blob:) fallback in 3 Uppy uploaders (follow-up to #292)

fabiodalez-dev:
## Summary

Follow-up to #292 (hero upload under a strict CSP). A CSP compatibility audit of the frontend found that the exact `fetch(blob:)` class the #292 fix addressed still lives as a **fallback branch** in three Uppy uploaders:

- `app/Views/cms/edit-home.php` — hero background (the file #292 touched: the fix reordered the branches but left the fallback)
- `app/Views/settings/index.php` — logo
- `app/Views/events/form.php` — event image

`file.preview` can be a `blob:` URL Uppy generates for image thumbnails, so `fetch(file.preview)` violates a strict `connect-src 'self'` policy (no `blob:`) — the same bug as #292.

## Why it's low-risk / not exploitable today

Every uploader in the app uses only `UppyDragDrop` — no remote-source plugin (Url/Webcam/GoogleDrive/…) is `.use()`d anywhere. With local drag-drop/file-input, `file.data` is always a real `File`/`Blob`, so the `instanceof File` / `instanceof Blob` branches always win and the `fetch(file.preview)` fallback is **dead code**. It would arm the moment a remote-source plugin is added — hence closing the latent bug class rather than just the reachable instance.

## Change

Removed the `fetch(file.preview)` fallback in all three uploaders. The File/Blob branches cover every case `UppyDragDrop` produces; anything else fails closed instead of doing a `blob:` fetch. `edit-home` also gains the `instanceof Blob` branch the other two already had, for symmetry. This matches the fetch-free pattern already used by `autori/crea_autore.php` and `libri/partials/book_form.php` (`new File([file.data])` + `FileReader`).

## Testing

- `php -l` clean on all three views.
- The #292 E2E (`tests/hero-upload-292.spec.js`) still passes: hero upload works and fetches no `blob:` URL.
- The rest of the frontend was audited and confirmed CSP-compatible: 137 `fetch()` calls all same-origin (`'self'`), no external-origin fetch/XHR/WebSocket, and the remaining `createObjectURL` uses are `img.src` previews or file-download clicks (governed by `img-src`/navigation, not `connect-src`).

### DIFF
--- a/app/Views/cms/edit-home.php
+++ b/app/Views/cms/edit-home.php
@@ -739,17 +739,15 @@ document.addEventListener('DOMContentLoaded', function() {
             if (file.data instanceof File) {
                 dataTransfer.items.add(file.data);
                 fileInput.files = dataTransfer.files;
-            } else if (file.preview) {
-                fetch(file.preview)
-                    .then(res => res.blob())
-                    .then(blob => {
-                        const newFile = new File([blob], file.name, { type: file.type });
-                        dataTransfer.items.add(newFile);
-                        fileInput.files = dataTransfer.files;
-                    })
-                    .catch(err => {
-                        console.error('Error converting file:', err);
-                    });
+            } else if (file.data instanceof Blob) {
+                // file.data is a Blob but not a File — wrap it directly. No
+                // fetch(file.preview): file.preview can be a blob: URL, and
+                // fetching it violates a strict connect-src CSP (the #292 bug
+                // class). With only UppyDragDrop in use, file.data is always a
+                // File/Blob; a non-Blob file.data (only from a remote-source
+                // Uppy plugin, none used here) fails closed instead.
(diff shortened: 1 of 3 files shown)


Himura2la:
FYI: hope this fixes the remaining upload issue https://github.com/fabiodalez-dev/Pinakes/issues/292#issuecomment-5088809623

fabiodalez-dev:
Superseded by #297. That PR removes the same dormant `fetch(file.preview)` fallbacks in edit-home, events and settings — and goes further: it wires Uppy's native ThumbnailGenerator preview into the hero and event uploaders, keeps the logo's object-URL preview (SVG-safe), and adds a browser test that every uploader shows a preview and transfers the file to the hidden input. Closing this to avoid a merge conflict; the cleanup ships in #297.

PR #298, opened 2026-07-27
Remove min-height from book title style

Himura2la:
Before 
Screenshot 2026-07-27 184920

After
image

## Summary by CodeRabbit

* **Correzioni**
  * Migliorata la visualizzazione delle card dei libri su dispositivi mobili, mantenendo un’altezza minima coerente per i titoli anche su schermi più piccoli.

### DIFF
--- a/app/Views/frontend/catalog-grid.php
+++ b/app/Views/frontend/catalog-grid.php
@@ -275,7 +275,6 @@ $getBookStatusBadge = static function ($book) {
     
     .book-title {
         font-size: 1rem;
-        min-height: 2.4em;
     }
 }
 


Himura2la:
Thanks to Custom CSS working on OPAC pages, I was able to test this xD

fabiodalez-dev:
The alignment problem you're pointing at is real. Digging in, though, this specific change doesn't fix it: with `.book-title { min-height: 2.8em }` still global, removing the mobile `2.4em` override just makes the mobile title inherit the taller base — and it doesn't touch what actually misaligns the rows (the optional subtitle, which unlike author/publisher didn't reserve its space).

I've opened #302, which keeps the intent — cards that line up — and makes it work: the subtitle stays conditional (no wasted space on the rows without one), and a per-row script equalises title and subtitle heights so any row that has a subtitle lines up across all its cards, while rows without one stay compact. Closing this in favour of #302.

PR #312, opened 2026-08-01
fix: recover from the maintenance-mode lockout after an update

fabiodalez-dev:
## Why

An in-app update writes `storage/.maintenance` and the front controller blocks every request while the file exists. Two independent flaws could leave a site permanently stuck in maintenance — one of them hit a production install after a half-failed update, with no way back in.

### 1. Login was blocked, but recovery needs an admin (catch-22)

The maintenance gate blocked the login route, yet the emergency "clear maintenance" endpoint is admin-only. A logged-out admin could never authenticate to reach it. The site was locked, and the one account that could unlock it was locked out too.

### 2. The flag was only cleared on a fatal error

Both `register_shutdown_function` handlers in the updater removed `.maintenance` **only** when a fatal PHP error occurred. A non-fatal early return — or a caught exception that skipped `disableMaintenanceMode()` — left the flag behind until the 30-minute stale sweep.

## The fix — three layers

- **Front controller (`public/index.php`)** — allow-list the login/logout routes (localised slugs + literal defaults) so an admin can always authenticate, and allow only the scoped `/admin/updates` UI/API needed to drive or abort the update. Privileged sessions do not bypass maintenance for arbitrary routes. The session is read read-only (`read_and_close`) with `session.save_path` pointed at `storage/sessions` — the same path the app configures later — otherwise the probe reads an empty session from `/tmp` and the bypass never fires.
- **Updater (`app/Support/Updater.php`)** — both shutdown handlers clear `.maintenance` for the owning request. Normal cleanup disarms the fallback before releasing `flock`, and the shared lock inode remains persistent to avoid cross-request unlink races. The update runs entirely within one request (`set_time_limit(0)` + `ignore_user_abort`), so once the request ends the update is over regardless of how it ended. Fatal errors are still logged for diagnostics.
- **Stale sweep** — the existing 30-minute stale-file check stays as the last resort for a SIGKILL/OOM that skips shutdown handlers entirely.

## Also: the maintenance page now matches the app

The 503 page was ours but off-brand (a purple gradient, system fonts, an indigo spinner). Rebuilt in the editorial-minimal identity — Fraunces serif wordmark and heading, Instrument Sans body, the `#d70161` brand accent, a clean card, a dark-mode variant, and reduced-motion support. Self-hosted fonts load from the already-allow-listed `/assets/`, with system fallbacks if the assets are mid-swap.

## Tests

`tests/maintenance-recovery-lockout.unit.php` (11 checks): behavioural HTTP checks with the flag present — login reachable, unauthenticated `503`, update/asset allow-list — plus source guards for the scoped recovery allow-list, ownership handoff and race-free cleanup. The admin-bypass path was verified end-to-end through a real browser login (which is how the `save_path` mismatch surfaced — a pure source guard would have missed it). PHPStan level 5 clean.

## Summary by CodeRabbit

* **Nuove funzionalità**
  * La modalità manutenzione consente login e logout, inclusi gli URL localizzati.
  * Amministratori e operatori autenticati possono accedere durante la manutenzione.
  * Nuova pagina 503 responsive con branding configurabile, logo personalizzabile, tema scuro e supporto alle installazioni in sottocartella.

* **Correzioni**
  * Migliorata la pulizia dei file temporanei al termine degli aggiornamenti, evitando interferenze tra richieste concorrenti.

* **Test**
  * Aggiunti controlli di regressione per accesso, aggiornamenti e ripristino dalla modalità manutenzione.

### DIFF
--- a/app/Support/BackupManager.php
+++ b/app/Support/BackupManager.php
@@ -324,16 +324,12 @@ class BackupManager
         // Safety net: a fatal error mid-restore must not leave the site locked
         // in maintenance (the front controller's 30-minute staleness fallback
         // remains the last resort). Mirrors Updater's shutdown handler.
-        register_shutdown_function(static function () use ($maintenanceFile, $lockFile): void {
+        register_shutdown_function(static function () use ($maintenanceFile): void {
             $error = error_get_last();
             if ($error !== null && in_array($error['type'], [E_ERROR, E_PARSE, E_CORE_ERROR, E_COMPILE_ERROR], true)) {
                 if (file_exists($maintenanceFile)) {
                     // nosemgrep: php.lang.security.unlink-use.unlink-use -- internal maintenance flag under storage/, not user input
                     @unlink($maintenanceFile);
-                }
-                if (file_exists($lockFile)) {
-                    // nosemgrep: php.lang.security.unlink-use.unlink-use -- internal lock file under storage/cache, not user input
-                    @unlink($lockFile);
                 }
             }
         });
@@ -356,10 +352,8 @@ class BackupManager
             }
             flock($lockHandle, LOCK_UN);
             fclose($lockHandle);
-            if (file_exists($lockFile)) {
-                // nosemgrep: php.lang.security.unlink-use.unlink-use -- internal lock file under storage/cache, not user input
-                @unlink($lockFile);
-            }
+            // Keep the shared lock inode persistent. Unlinking after unlock can
+            // delete a subsequent updater/restore request's lock path.
         }
     }
 
--- a/app/Support/Updater.php
+++ b/app/Support/Updater.php
@@ -1745,7 +1745,19 @@ class Updater
         ]);
 
         $maintenanceFile = $this->rootPath . '/storage/.maintenance';
-        register_shutdown_function(function () use ($maintenanceFile, $lockFile) {
+        // Ownership flag: only the request that actually acquires the update lock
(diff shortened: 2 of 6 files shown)


fabiodalez-dev:
Addressed the review in 76e73700:

- **Lock ownership (Critical)** — both shutdown handlers now bail unless this request owns the update lock. An ownership flag is raised only after `flock()` succeeds and maintenance is on, so a request rejected at `flock` (another update already running) no longer tears down the active update's `.maintenance`/lock. The owner still clears them unconditionally, preserving the recovery behaviour.
- **Session probe (Major)** — the probe now runs only when a session cookie is already present (anonymous 503 traffic creates nothing), in `use_strict_mode`, with the app's cookie flags applied up front so any emitted cookie is `HttpOnly`/`Secure`/`SameSite`. I did **not** apply the suggested `use_cookies=false`: with it PHP stops reading the id from the incoming cookie, so the admin's session becomes invisible and the bypass never fires — I verified this against a real browser login (admin saw the 503 page). Reading stays enabled; the hardening targets what the probe may emit.
- **Test check 01 (Trivial)** — the "reachable" checks now require an actually-served 2xx/3xx, not merely "not 503", so a 404/500 fails them.

The test also gained source guards for the ownership flag on both handlers and the cookie-gated strict-mode probe. Admin bypass re-verified end-to-end; anonymous visitors get 503 with no `Set-Cookie`. PHPStan level 5 clean, 14/14 checks pass.

PR #324, opened 2026-08-04
perf + data-integrity: cut TTFB/DB load and harden import/export, copies, availability and interop

fabiodalez-dev:
This bundles two lines of work — a performance pass and a data-integrity audit — plus the fixes from the automated review.

## Performance: cut TTFB and per-request DB load

Bootstrap (every request):
- `ConfigStore`: cache raw `system_settings` rows across requests (APCu/file, 60s TTL + explicit invalidation on every write path) instead of opening a second MySQL connection and re-reading the whole table per request; reuse one shared connection for multi-key admin saves.
- `PluginManager`: run bundled-plugin sync and orphan cleanup at most once per 15 minutes (re-triggered on release or plugin lifecycle actions), cache active plugins + hook rows in one payload, prefetch hooks with a single query instead of one per plugin.
- `ThemeManager`/`I18n`/`SchemaInfo`: cache the active theme row, the active languages list, and positive table-existence probes (all invalidated on the relevant mutations).
- Digital Library / GoodLib plugins: stop re-upserting their hook rows on every request — self-heal only when rows are missing.

Pages and queries:
- Home: build the whole visitor-independent dataset once (single `home_content` read instead of four queries, latest books, genre carousels, events, hero counters) and cache it for 5 minutes; render the hero counters and latest-books grid server-side (no spinners + two API round-trips before first paint); preload the hero background image (LCP).
- Catalog: cache total/availability counts and the six sidebar facet aggregations per filter set (2 min TTL, cleared on book changes).
- New composite indexes on `libri` for the newest-first sorts — `(deleted_at, created_at)` and `(genere_id, deleted_at, created_at)` — shipped in `schema.sql` + `migrate_0.7.53.sql` with a behavioural test.
- composer: enable `optimize-autoloader` + `apcu-autoloader`.

Measured on a seeded install: warm home page 91 → 17 queries per request, catalog API TTFB −42%.

## Data-integrity audit

An extreme data-consistency pass across the whole read/write surface, so the same book data reads the same everywhere and no routine operation loses data.

- **Copies & availability**: the edit form no longer hard-deletes physical copies when a book has any out-of-circulation copy (in-circulation count now used consistently on both the pre-validation and the copy-management paths); create runs the availability recalc like update; the web loan-request gate goes through `CapacityService` like the mobile app, and `calculateAvailability` + `DataIntegrity` + the notification/overbooked queries all use the same `COALESCE(data_inizio_richiesta, DATE(data_scadenza_prenotazione))` start; the mobile API classifies `da_ritirare` (and a pending conversion loan holding a copy) as reserved, not on-loan.
- **CSV / LibraryThing round-trip**: export the `co_autori`, `illustratore`, `curatore`, `colorista` and `classificazione_dewey` columns (raising `group_concat_max_len`); import splits a multi-publisher cell into primary + co-publishers (#143), creates genres named in the CSV, keeps the first non-empty value on duplicate-mapped headers, normalizes the free-text contributor columns, keeps `descrizione_plain` in step and preserves rich HTML on a pure round-trip, reverses formula-injection escaping, raises the copies cap, and never wipes existing authors when the author cell is empty; contributor roles round-trip through distinct columns instead of promoting co-authors to principal.
- **Genres, locations, facets, interop**: genre merge/delete no longer destroys shelf positions or unshelves books; author merge repoints authority/import-source rows; publisher facet groups by name; author book counts are DISTINCT; OPAC cards fall back to co-author and show the publisher; Z39.50/SRU, NCIP and the public API report availability from the canonical counters, resolve shelf location from the current columns, absolutize cover URLs, and export secondary publishers, legacy contributors and pseudonyms.

## Review fixes

The automated review's actionable findings are addressed (soft-delete guard + `prepare()` checks on the description lookup, a `descrizione_plain` column-existence guard for pre-migration schemas, `prepare()` validation + statement close in the genre position-reassignment, canonical holdings counts in the MODS/MARCXML formatters, `htmlspecialchars` in the home badges, and a single `DescriptionText::toPlain()` source of truth). A few are intentionally deferred with reasons noted in the review thread.

## Tests

- `tests/data-integrity-audit.spec.js`: 21 E2E tests, one per fixed point, driving the real app and asserting against the live DB.
- Unit coverage for the reservation start fallback, contributor role round-trip, genre cascade delete, author-merge integrity, and the performance cache regressions.

Full app-tree PHPStan is clean; the E2E and unit suites are green.

## Summary by CodeRabbit

- **Nuove funzionalità**
  - Importazione ed esportazione CSV ampliate con coautori, traduttori, illustratori, curatori e coloristi.
  - Le API e i cataloghi bibliografici esterni espongono informazioni più complete su contributori, editori, disponibilità e stati dei libri.
  - La homepage mostra più contenuti immediatamente e supporta caricamenti più efficienti.

- **Miglioramenti**
  - Homepage, ricerca e filtri risultano più rapidi grazie a ottimizzazioni delle prestazioni.
  - Migliorata la gestione di disponibilità, prenotazioni legacy, copie non circolanti e collocazioni.
  - Corretti duplicati nei conteggi di libri e autori e rafforzata l’integrità dei dati importati.


fabiodalez-dev:
Thanks for the review — here's how I handled each actionable finding. (This PR supersedes #323; the fixes below are already in this branch.)

### Addressed

- **LibriController — copy-reduction pre-validation**: now uses `countInCirculationByBookId()` too, so it no longer diverges from the copy-management count and wrongly blocks a save when out-of-circulation copies exist.
- **Mobile API — pending conversion loan**: a `pendente` loan already holding a copy (`attivo=0`, `copia_id NOT NULL`) is now classified as `reserved`, matching `CapacityService`/`DataIntegrity`, instead of falling through to `unavailable`.
- **MARCXML 866 available count**: falls back to counting copy rows (not `0`) when the canonical counter is absent.
- **MODS holdings summary**: now reads the canonical `copie_totali`/`copie_disponibili` counters, like MARCXML.
- **CsvImportController — description lookup**: added `AND deleted_at IS NULL` and a `prepare()` false-check.
- **CsvImportController — `descrizione_plain`**: written behind a column-existence guard (pre-migration schema safety), mirroring the LibraryThing importer.
- **GenereRepository — position reassignment**: `prepare()` return values are validated and the four statements are closed after the loop.
- **DescriptionText**: `LibriController::normalizeDescriptionForCsv()` now delegates to `DescriptionText::toPlain()` (single source of truth; the CR-strip step moved into the shared helper).
- **home-books-grid.php**: status badges use `htmlspecialchars(..., ENT_QUOTES, 'UTF-8')` instead of `HtmlHelper::e()`.
- **data-integrity-audit.spec.js**: the skip guard also requires the DB env the direct DB helpers use; the unavailable-badge test lists the book via an `editore` filter (deterministic) instead of the FULLTEXT `q` search and asserts the visible "Non disponibile" label.

### Intentionally deferred (with reasons)

- **CollocazioneController — centralize the fallback**: the inline fallback already emits the canonical dash-padded format; `CollocationRepository::buildCollocazioneString()` has a different ID-based interface, so unifying them is a separate refactor rather than a correctness fix here.
- **CsvImportController — copy-creation query volume**: the higher cap protects against DoS while letting a real cross-install migration keep its copies; batch inserts are a larger change and the per-copy loop only bites on pathological rows, so I've left it for a dedicated pass.
- **LibriController — `GROUP_CONCAT(DISTINCT … ORDER BY)`**: the names are the distinct values and `a.nome` is in the `ORDER BY`, so the output is stable; `ordine_credito` is a best-effort secondary matching the prior behaviour. Not worth a derived-table rewrite here.
- **ReservationsController — batch the conflict-date probe**: it only runs on the error path after the whole-range check already failed, and is bounded by `max_loan_duration_days`, so the added query volume is limited.
- **LibraryThingImportController — legacy contributor writers**: these columns are a denormalized cache; the importer maintains the authoritative entity links (`libri_autori`) correctly and the export prefers entities over the cache, so aligning the cache writers is a focused LT change better done on its own.
- **GoodLibPlugin — `commit()`/`RELEASE SAVEPOINT` return checks**: outside the scope of these changes (pre-existing plugin code, not touched by this work).

PHPStan is clean on the full app tree; the 21 E2E tests and the unit suites are green.

fabiodalez-dev:
All actionable findings from this pass are addressed in 8517f902 — nothing deferred this time.

**Correctness / data integrity**
- `GenereRepository::delete()` now runs the usage checks + DELETE in one transaction with a `FOR UPDATE` lock (no TOCTOU cascade of `posizioni`); `cascadeDelete()` checks `execute()` before `commit()`; the position move excludes soft-deleted books (`deleted_at IS NULL`).
- `LibriController`: the copy-reduction validation reports the in-circulation floor (`current − removable`); the selected-export refuses with a message instead of shipping a partial CSV above 1000 IDs.
- `CsvImportController`: the existing-description lookup selects `descrizione_plain` only when the column exists (pre-migration safety).
- `AuthorRepository`: the `author_authority_alternates` / `libri_autori_import_sources` repoints are gated on table existence.

**Performance**
- `CsvImportController` creates copies with one prefix pre-load + one batch INSERT (`CopyRepository::createManyForBook`) instead of two queries per copy.
- `CapacityService::unavailableDatesInRange` loads the intervals once for the whole span; `ReservationsController::capacityConflictDates` uses it (no per-day query storm).
- `DataIntegrity` batched recalc invalidates the book caches once per chunk.
- SRU matches secondary publishers via a correlated `EXISTS` instead of the fan-out `LEFT JOIN` on `libri_editori`.
- `Language` invalidates `i18n_languages` once, only after the final consistent state; event-settings save refreshes the home cache.

**Robustness / maintainability**
- Digital Library & GoodLib: the per-request hook self-heal catches `\Throwable` (activation still propagates); `commit()`/`RELEASE SAVEPOINT` return values are checked.
- LibraryThing export guards the query and returns a clean error rather than a malformed file.
- `CollocazioneController` has a single collocazione-fallback helper; `formatCopyStatus` is centralised in `RecordFormatter`; MARCXML/MODS holdings read the canonical counters; the migration's `WHERE` is on its own line.

**Tests**
- `data-integrity-audit`: quote-aware CSV parsing, run-unique author fixtures, DB env in the skip guard, badge asserted via a deterministic `editore` filter.
- `contributor-roles-237` creates both roles locally; `plugin-hook-transaction-contract` resolves `E2E_DB_*` creds and derives the expected hook count; `performance-cache-regressions` uses method-existence checks instead of source-text matching; `issue-255` waits for the settings redirect.

PHPStan is clean on the full app tree; the 21 E2E tests and the unit suites (contributor-roles 51, pr323 105, plugin-hook 16, perf-cache 49, genre-cascade 11, author-merge 9, loan 9+, migration 6, contributor-roundtrip 18) are green.

fabiodalez-dev:
This review ran against `8517f90`; every finding here is already resolved in the following commit (`d8e906c6`), which is now the branch head:

- `CopyRepository::createManyForBook` checks both `prepare()` results and closes the statement on a failed `execute()`.
- `Language::create()/update()` wrap the row write and the default switch (`setDefaultRows()`) in one transaction and invalidate `i18n_languages` once after commit — no observable multiple-default window.
- SRU `!=` treats a NULL column as non-matching: `({col} IS NULL OR {col} NOT LIKE …)`, so a book with only a co-publisher is no longer excluded from `dc.publisher != …`.
- `RecordFormatter` merges legacy contributors by normalized name AND role (a second legacy name is kept even when the first is already an entity).
- The self-heal savepoints/probes in Digital Library & GoodLib now use per-invocation random names, so they can't clobber a caller savepoint (covered by new cases in plugin-hook-transaction-contract).
- `issue-255` waits on the POST response to `/admin/settings/registration` instead of `waitForURL`.

PHPStan is clean; E2E (21) and the unit suites are green.

fabiodalez-dev:
Shipped in v0.7.53 via the release integration branch (release/0.7.53-rc.2 → main). Closing as delivered.

PR #332, opened 2026-08-05
feat(i18n): non-reseeded override layer for in-app locale edits (@48)

fabiodalez-dev:
## What

Review walkthrough **@48 (Option B)**: the Docker image re-seeds `locale/*.json` on every boot, so an in-app edit of a shipped translation (written straight to `locale/{code}.json`) was **silently reverted on restart**. This adds a non-reseeded **override layer** so those edits persist while shipped-translation updates still reach existing installs.

## How

- **`I18n::loadTranslations()`** now loads the shipped base `locale/{code}.json` and then merges `locale/overrides/{code}.json` on top (override wins). Keys shipped only in a later base image still appear (they're absent from the override). No override → byte-for-byte the old behaviour.
- **`LanguagesController`**:
  - the in-app editor writes the edit to `getOverrideFilePath()` (`locale/overrides/{code}.json`), with the path-traversal guard retargeted at that subdir;
  - **download** serves the *effective* (base+override) set, so a re-download reflects what the app shows and a re-upload never drops prior overrides;
  - **delete** removes both base and override so a re-added code doesn't resurrect a stale override.
  - the canonical key list (`loadCanonicalTranslations`, `it_IT`) stays the shipped base.

Pairs with pinakes-docker `feat/locale-override-dir` (ensures `locale/overrides/` exists; README updated).

## Tests

- `tests/locale-override-layer.unit.php` drives the **real** `I18n` loader against sandbox base+override files (override wins; base-only key from base; override-only key applied; removing the override falls back to the shipped base) plus source invariants of the editor write/delete/download.
- `discussion-238-followup` export assertion updated to the equivalent new shape.
- PHPStan level 5 clean (full tree); full unit suite green; locale parity unchanged (6765 keys).

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Aggiunto un livello di override per personalizzare le traduzioni senza modificare i file di base.
  * Le traduzioni personalizzate prevalgono automaticamente sui valori predefiniti.
  * Il download esporta le traduzioni effettivamente applicate, incluse le chiavi aggiuntive.

* **Correzioni**
  * Migliorata la gestione delle traduzioni mancanti e dei file non validi.
  * L’eliminazione di una lingua rimuove anche le personalizzazioni associate, conservando backup separati.

### DIFF
--- a/app/Controllers/Admin/LanguagesController.php
+++ b/app/Controllers/Admin/LanguagesController.php
@@ -293,12 +293,15 @@ class LanguagesController
         try {
             $languageModel->delete($code);
 
-            // Optionally delete translation file
-            $translationFile = $this->getLocaleFilePath($code);
-            if (file_exists($translationFile)) {
-                // Backup before deleting
-                copy($translationFile, $translationFile . '.deleted.' . time());
-                unlink($translationFile);
+            // Optionally delete translation file(s): the shipped base and the
+            // admin override, so a deleted language leaves no lingering override
+            // that would resurface if the same code is re-added later.
+            foreach ([$this->getLocaleFilePath($code), $this->getOverrideFilePath($code)] as $translationFile) {
+                if (file_exists($translationFile)) {
+                    // Backup before deleting
+                    copy($translationFile, $translationFile . '.deleted.' . time());
+                    unlink($translationFile);
+                }
             }
 
             $_SESSION['flash_success'] = __("Lingua eliminata con successo");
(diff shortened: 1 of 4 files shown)


fabiodalez-dev:
Merged into main and shipped in **v0.7.55**; the three CodeRabbit findings on this review are fixed in **v0.7.56** (commit 159e33c4):
- download()/refreshStats() now use `getEffectiveTranslations()` and accept base **or** override existence (override-only new locales work);
- delete() checks `unlink()` and flashes a warning on a residual file instead of silently claiming success;
- the override unit test saves & restores any pre-existing `zz_ZZ` catalog.
PHPStan L5 clean, full unit suite green. Closing as delivered.

PR #353, opened 2026-08-13
fix(scrape): drop deprecated curl_close() calls that corrupt ISBN-import JSON on PHP 8.5

fabiodalez-dev:
## Summary

Fixes the ISBN import failing with **"Risposta non valida dal servizio ISBN."** on PHP 8.5, even though the scrape itself succeeds.

`curl_close()` has been a no-op since PHP 8.0 (curl handles are objects, freed by the GC) and emits a `Deprecated` notice on PHP 8.5+. With `display_errors` on, that notice is written to the response **body before the JSON payload**, so the browser's `response.json()` throws and the form shows the "invalid response" error. The z39-server SBN/SRU clients sit in the ISBN scrape path, which is where it first bit.

Reproduced against the real PHP-FPM 8.5 request (`GET /api/scrape/isbn?isbn=8804763175`): the body was prefixed with several
`Deprecated: Function curl_close() is deprecated since 8.5 … in storage/plugins/z39-server/classes/SbnClient.php on line 957`
HTML notices, corrupting an otherwise-complete JSON payload.

## What changed

Removed every active `curl_close($handle)` across `app/` and the bundled plugins (z39-server, open-library, book-club) plus core (`Updater`, `CoverController`, `LibriController`, `SsrfGuard`, `web.php`), replacing each with the marker comment already used in the `scraping-pro` plugin:

```php
/* curl_close(): no-op since PHP 8.0, deprecated 8.5 */
```

After the fix the same request returns clean `application/json` (`title: "La vedova"`, `isbn13: 9788804763178`, `dewey: 813.6`), no `Deprecated` in the body.

## Regression guard

`tests/no-deprecated-curl-close-php85.unit.php` scans the shipped code and fails if any active `curl_close($handle)` call reappears (the retained marker comment uses empty parens, so it is never matched).

## Summary by CodeRabbit

* **Bug Fixes**
  * Migliorata la compatibilità con PHP 8.5 rimuovendo l’uso di funzioni obsolete nella gestione delle richieste HTTP.
  * Mantenuti invariati il recupero, la validazione e l’elaborazione di immagini, copertine e risposte dei servizi esterni.

* **Tests**
  * Aggiunto un controllo automatico per individuare eventuali utilizzi residui di funzioni non più supportate.

### DIFF
--- a/app/Controllers/CoverController.php
+++ b/app/Controllers/CoverController.php
@@ -353,7 +353,7 @@ class CoverController
             $rawResponse = curl_exec($ch);
             if ($rawResponse === false) {
                 $err = curl_error($ch) ?: 'Errore sconosciuto';
-                curl_close($ch);
+                /* curl_close(): no-op since PHP 8.0, deprecated 8.5 */
                 throw new \RuntimeException('Errore download immagine: ' . $err);
             }
 
@@ -361,7 +361,7 @@ class CoverController
             $headerSize = (int) curl_getinfo($ch, CURLINFO_HEADER_SIZE);
             $headers = substr($rawResponse, 0, $headerSize);
             $body = substr($rawResponse, $headerSize);
-            curl_close($ch);
+            /* curl_close(): no-op since PHP 8.0, deprecated 8.5 */
 
             if ($status >= 300 && $status < 400) {
                 if ($redirects === 3) {
--- a/app/Controllers/LibriController.php
+++ b/app/Controllers/LibriController.php
@@ -2084,7 +2084,7 @@ class LibriController
             curl_exec($ch);
             $contentLength = curl_getinfo($ch, CURLINFO_CONTENT_LENGTH_DOWNLOAD);
             $httpCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
-            curl_close($ch);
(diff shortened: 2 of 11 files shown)


fabiodalez-dev:
Both review comments are addressed in 2647de17 (pushed after the review ran):

- **Explicit TLS verification** — `SbnClient` (parallel + `makeRequest`), `SoggettarioClient` and `open-library`'s `checkCoverExists()` now set `CURLOPT_SSL_VERIFYPEER => true` / `CURLOPT_SSL_VERIFYHOST => 2` instead of relying on the runtime default, and `checkCoverExists()` also gained `CURLOPT_PROTOCOLS`/`REDIR_PROTOCOLS`/`MAXREDIRS`/`CONNECTTIMEOUT` plus a `curl_error()` check after each `curl_exec()`.
- **Multi-line detection in the guard test** — `tests/no-deprecated-curl-close-php85.unit.php` now tokenises each file with `token_get_all()` instead of a line-based regex, so a call split across lines is caught; verified it flags single-line and multi-line calls while ignoring method calls, the marker comment and empty-argument calls.

fabiodalez-dev:
@coderabbitai review

PR #364, opened 2026-08-19
fix(plugins): run the lifecycle on ZIP-update of an active plugin (0.7.62 blocker)

fabiodalez-dev:
Fixes the plugin-ZIP-update lifecycle blocker discovered on the 0.7.62 release branch (#363).

## The bug

`PluginManager::updatePluginFromStaging()` swapped the plugin's files and saved the new version immediately, but for an **active** plugin it never re-ran `onActivate()` / `ensureSchema()` nor registered the plugin's new hooks. An update that adds a column, table or callback therefore reported success while leaving the plugin in an inconsistent state — and because the stored version was already bumped, the next boot saw "already up to date" and never healed it.

## The fix

- The active-plugin ZIP update now reconciles the lifecycle so a schema/hook-changing update actually applies (or the plugin is left in a state the next boot repairs), instead of a silent half-update.
- New upgrade integration coverage that changes **both** the schema and the registered hooks across the update, asserting the post-update state — the gap my earlier `plugin-zip-update-all-bundled` test missed (it only checked id/files/version).
- Release-packaging hardening (`.distignore`, `.rsync-filter`, `bin/build-release.sh`, `scripts/ci-verify-release.sh`) so packaging sessions stay reproducible.

Targets the `release/0.7.62-rc.1` branch so the fix ships inside #363 (the 0.7.62 release PR).

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Gli aggiornamenti dei plugin attivi vengono completati in modo sicuro al riavvio, con backup e ripristino automatico in caso di errore.
  * Gli aggiornamenti non riusciti ripristinano file, metadati e hook precedenti.

* **Correzioni**
  * I pacchetti di distribuzione escludono i dati delle sessioni runtime, mantenendo la struttura necessaria.
  * I plugin non ripristinabili vengono temporaneamente esclusi per evitare ulteriori errori.


fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
Addressed in 92898b95:

- **PluginManager.php (Major — unreadable marker bricks updates):** an unreadable marker is now retired — renamed to a non-\`.json\` name for diagnosis, unlinked if the rename fails — so a corrupt marker can no longer make every future update fail with "already pending" via \`fopen('x+b')\`.
- **build-release.sh + ci-verify-release.sh (Major — inconsistent session-tree check):** both verifiers now require \`storage/sessions/.gitkeep\` and reject every other entry (files, symlinks, stray dirs) via \`-mindepth 1 ! -name .gitkeep\`.
- **plugin-manager.unit.php (Minor — tautological assertions):** the \`onActivate\` assertion is anchored inside \`finalizePendingPluginUpdate\`, and the ordering check now compares real positions (offset 0). Added two regression guards for the marker retirement and the non-throwing backup cleanup.

Also hardened \`deletePendingPluginBackup()\` to log-and-continue instead of throwing after a committed update — a throw there propagated into the finalize catch and rolled back an already-applied, already-activated update.

PR #365, opened 2026-08-19
Release 0.7.62: apply plugin-update lifecycle blocker fix (#364) to main

fabiodalez-dev:
Brings the plugin-update **lifecycle blocker fix (#364)** into `main` so the
`v0.7.62` tag ships it.

## Why a second PR
The 0.7.62 release PR merged to `main` at head `57baf0d0`, one step before the
active-plugin ZIP-update lifecycle fix landed. That fix (and its follow-up
hardening) went into `release/0.7.62-rc.1` afterwards, so `main` is currently at
0.7.62 **without** the blocker fix. This PR is the exact delta between `main`
and the release branch — nothing else.

## What it contains
- **Active-plugin ZIP-update lifecycle** (`746c9131`): updating an already-active
  plugin now runs its new `onActivate()`/`ensureSchema()` on the next request via
  a pending-update marker, rolling back package, metadata and hooks if the new
  version fails to activate — so schema changes shipped in an update are applied
  instead of being silently skipped.
- **Hardening** (`92898b95`): an unreadable update marker is retired instead of
  permanently blocking future updates (`fopen('x+b')` would otherwise fail with
  "already pending" forever); `deletePendingPluginBackup()` logs-and-continues
  instead of throwing after a committed update; both release verifiers require
  `storage/sessions/.gitkeep` and reject every other entry (files, symlinks,
  stray dirs); two tautological unit assertions were rewritten to actually
  verify what they claim, plus regression guards for the two fixes above.
- **CHANGELOG** (`226df253`): document the lifecycle and verifier changes under
  0.7.62.

## Verification
The delta already went green on the full CI (33/33) and was reviewed by
CodeRabbit on the release branch (8/8 threads resolved). Plugin tests locally:
`plugin-zip-update.integration` PASS, all-bundled 20/20, `plugin-manager.unit`
23/23, PHPStan clean.

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Gli aggiornamenti dei plugin attivi completano il ciclo al successivo avvio, con ripristino automatico in caso di errore.
  * Aggiunto l’invio di solleciti per singoli prestiti o per più prestiti selezionati.

* **Miglioramenti**
  * I solleciti sono disponibili solo per prestiti attivi e scaduti.
  * Migliorata la gestione di prestiti, prenotazioni, rinnovi e notifiche in situazioni concorrenti.
  * I pacchetti escludono i dati runtime delle sessioni, mantenendo la struttura necessaria.

* **Correzioni**
  * Rafforzati i controlli per evitare doppie assegnazioni e notifiche duplicate.
  * I marcatori di aggiornamento illeggibili vengono ritirati senza bloccare gli aggiornamenti successivi.

* **Test**
  * Ampliate le verifiche su plugin, pacchetti e flussi di prestito e prenotazione.

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -75,6 +75,7 @@ fix-*.php
 # Storage/uploads (exclude user data, keep directory structure)
 storage/logs/*
 storage/cache/*
+storage/sessions/*
 storage/backups/*
 storage/uploads/*
 storage/tmp/*
@@ -87,6 +88,7 @@ cache/*
 # Keep essential files and directory structure
 !storage/logs/.gitkeep
 !storage/cache/.gitkeep
+!storage/sessions/.gitkeep
 !storage/backups/.gitkeep
 !storage/backups/.htaccess
 !storage/uploads/.gitkeep
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,9 @@ storage/*
 !storage/logs/
 storage/logs/*
 !storage/logs/.gitkeep
+!storage/sessions/
+storage/sessions/*
+!storage/sessions/.gitkeep
 storage/*.log
 !storage/tmp/
 storage/tmp/*
--- a/.rsync-filter
+++ b/.rsync-filter
@@ -34,6 +34,8 @@
 + storage/logs/.gitkeep
 + storage/cache/
 + storage/cache/.gitkeep
++ storage/sessions/
++ storage/sessions/.gitkeep
 + storage/backups/
 + storage/backups/.gitkeep
 + storage/backups/.htaccess
@@ -254,6 +256,7 @@
 # Storage user data (structure already included above)
 - /storage/logs/*
 - /storage/cache/*
+- /storage/sessions/*
 - /storage/backups/*
 - /storage/tmp/*
 - /storage/calendar/*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,9 @@ Overdue-loan recalls (solleciti) and emailing the loan receipt (#360).
   single overdue notification. Automatic recalls repeat at a configurable
   interval up to a configurable cap (Settings → Loans → "Solleciti automatici",
   off by default; sent by the notifications cron or on admin login). Staff can
-  also send a manual recall for one loan from the loan detail page, or for many
-  at once from the loans list via the bulk action bar — manual recalls ignore
+  also send a manual recall for one loan — from the loan detail page or a
+  per-row action on the loans list — or for many at once from the loans list
+  via the bulk action bar — manual recalls ignore
   the automatic schedule but share the same per-loan counter
   (`prestiti.recall_count` / `last_recall_at`, added by
   `migrate_0.7.62-rc.1.sql` and self-healed at runtime). New editable email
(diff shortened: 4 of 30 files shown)


fabiodalez-dev:
All seven review findings addressed in bbaf1a0d:

- **PluginManager `skipPluginIdsThisRequest` (Major):** now `static`, so the skip covers every `PluginManager` instance in the request, not just the one that rolled back — the broken replacement class stays defined process-wide.
- **Fail-closed verifiers (Major):** `bin/build-release.sh` and `scripts/ci-verify-release.sh` now treat a non-zero `find` status as an error instead of "no session data" (the `|| true` / pipe-to-`grep` swallowed it).
- **restorePluginHooks transaction (Trivial):** DELETE + re-INSERT wrapped in a transaction, opened only when one is not already active (`@@autocommit` guard) so it never nests.
- **Unreadable/lock-failed marker (Minor):** now logs instead of persisting silently. It still retries rather than retiring the marker — retiring on a *transient* open/lock error would silently drop the deferred `onActivate` (the schema migration this whole mechanism exists to run); a genuinely permanent failure is now diagnosable in the log.
- **Duplicated marker prefix (Trivial):** `PENDING_UPDATE_PREFIX` is now `public`; the three test sites reference it.
- **stdout/stderr deadlock (Minor):** the fresh-bootstrap helper routes child stderr to a file, so a large stack trace can no longer deadlock the parent.
- **Third-bootstrap assertion (Minor):** it now asserts the rolled-back plugin state is *unchanged* by a later request (version/is_active/hooks), and the comment no longer claims to test the intra-process skip set, which a separate process cannot observe.

fabiodalez-dev:
Both findings on 0440150 addressed in f9656b1a:

- **restorePluginHooks @@autocommit detection (Trivial):** replaced with a savepoint probe. `@@autocommit` stays 1 after `begin_transaction()`, so it could not detect an outer transaction and would nest `begin_transaction()` (implicitly committing the caller's transaction). A `SAVEPOINT` outside a transaction is discarded by autocommit so its `RELEASE` fails — verified in/out of a transaction and after rollback. Inside a transaction the hook restore is now scoped to a `SAVEPOINT`/`ROLLBACK TO SAVEPOINT`; otherwise it owns a fresh transaction. Handles both mysqli exception and silent error modes.
- **Symlinked session dir/placeholder (Major):** both verifiers now reject a symlinked `storage/sessions` or `.gitkeep` with `[ -L ]` before scanning (`-d`/`-f` follow symlinks), allowing only a real directory holding a real placeholder. Verified behaviourally (real → pass, symlinked dir/placeholder → reject) and a static guard in `code-quality.spec.js` pins the `-L` check in both verifiers.

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
Addressed in 025e79c3:
- Both verifiers now reject a symlinked parent `storage/` too (`-L "$package_dir/storage"`), pinned by a static guard.
- `pickup-ready-copy-free-366.unit.php` reads `getenv(E2E_DB_*)` first and fails (not silently skips) under `CI_STRICT_TESTS`.
- The `runAll()` ordering check asserts the method was found before comparing positions (no `(int) strpos` false pass).
- Corrected the "touches only rows it creates" headers on the three tests that drive global maintenance sweeps — they mutate every matching row and must run against an isolated DB (as CI does).

PR #371, opened 2026-08-21
fix: localize gender labels in admin user details

fabiodalez-dev:
Supersedes #369 with the original contributor commit preserved, plus the review fixes required to make the change merge-ready. The external fork has maintainer edits disabled, so its branch could not be updated directly.

Changes:
- render stored `M`, `F`, and `Altro` values through the locale catalogue;
- preserve legacy decode-then-escape behavior for all existing detail fields;
- add a stable, locale-independent field hook;
- cover real registration persistence and all enum labels in the serial issue #255 browser suite.

Validation:
- issue #255 Playwright suite: 30/30 passed;
- PHP syntax, JS syntax, locale parity, Playwright policy, and diff checks passed.

Co-authored work from @Himura2la is preserved in the branch history.

PR #372, opened 2026-08-22
Release 0.7.64: multiple copies per borrower + circulation durability (#238)

fabiodalez-dev:
Consolidates the open #238 work and the two follow-up PRs into a single 0.7.64 release.

## What this bundles
- **#238 — multiple physical copies per borrower** (opt-in, off by default): staff can lend more than one distinct physical copy of the same title to one borrower; every loan stays bound to its own copy. Pending requests, reservations and legacy copyless rows keep the historical uniqueness. (Supersedes #368.)
- **Circulation durability** (from #370): the pickup-ready email becomes claim-and-retry via three new `prestiti` columns and an idempotent, RC-named migration; the DB triggers treat a stale overdue `in_corso` loan as open-ended; NCIP hardening (guarded cancel, closes a cross-user check-in hole) while staying strict; the release pipeline is split into a read-only build job and a privileged publish job with provenance.
- **Gender i18n on admin user details** (#371): the stored `sesso` enum renders through translated labels.
- **Mobile notifications dropdown** no longer clips on phones.

## Review fixes applied on top
- **F2**: approval and the date-aware allocator stay flexible for a copy that is physically out but whose open loan does not overlap the requested window (restored the pre-#370 behaviour; the per-copy overlap check remains the authority).
- **F1**: actively backdating an active loan's due date now returns a clear `expired_window` error instead of the opaque `loan_update_failed`, while still allowing edits to an already-overdue loan.
- Migration renamed `migrate_0.7.64-rc.1.sql` so the Updater selects it even when the release ships from a prerelease.

## Tests
- 25 new checks (`tests/loan-review-0764.unit.php` 19/19, `tests/loan-review-0764.spec.js` 6/6) against the real DB + triggers and the live browser, plus the existing multiple-copy E2E and migration units. PHPStan level 5 clean, i18n parity across all five locales.

Closes #368, #370, #371.


fabiodalez-dev:
@coderabbitai review

Issue #387, opened 2026-08-27
Performance: caching overhaul — measure, OPcache, QueryCache refactor, then optional Redis + edge cache


fabiodalez-dev:
## Goal

Make Pinakes substantially faster under real traffic, in an order driven by measurement rather than assumptions. The current cache stack (`QueryCache` = APCu + file fallback, `ContentCache` invalidation, OPcache) works but has concrete inefficiencies, and the biggest theoretical win (full-page/edge cache) is blocked by session + CSRF in the anonymous HTML. This tracks the full plan; the first PR delivers the safe, measured quick wins (steps 1–3).

## Principles / constraints

- **Existing installs first**: no mandatory migration, no forced infrastructure. Redis stays opt-in with a deterministic bootstrap choice; shared-hosting without Redis keeps working identically.
- **Soft-delete**: every cache key respects `deleted_at IS NULL`; invalidation already flows through `ContentCache`.
- **Per-user multilingual**: cache keys always include the locale; any future full-page cache is anonymous-only.
- **Coherence over blind TTLs**: prefer explicit write-path invalidation (already in `ContentCache`).

### 1. Measure first (baseline)
Instrument and capture, in production-like conditions: p50/p95 TTFB, SQL time & query count per request, cache **hit/miss ratio**, file-lock timeouts, OPcache/APCu saturation (`opcache_get_status()`, `apcu_sma_info()`). No optimization ships without a before/after number.

### 2. OPcache — safe and observable (no a-priori numbers, no JIT)
- Read `opcache_get_status()` in production before changing anything. `php.ini.recommended` already sets 128 MB / 10k files; the app is ~7k PHP files incl. vendor+plugins, so 128 MB may already suffice.
- Raise to 256 MB / 20k **only if** 128 MB is actually near saturation.
- Remove `opcache.fast_shutdown` (no-op since PHP 7.2).
- Do **not** set `validate_timestamps=0` until every code-mutating path invalidates OPcache in the right SAPI (manual updates, plugins, CLI deploys) — not only the Updater's `opcache_reset()` (`Updater.php:3750`). CLI OPcache is a separate cache from web/FPM.
- **No JIT by default**: the app is DB/I-O-bound; JIT helps CPU-bound PHP and is off by default on PHP 8.4. Measure separately if ever.
- APCu sizing from `cache_full_count` / `apcu_sma_info()`, not arbitrary numbers.

### 3. QueryCache refactor (the real first application win)
[...]

Issue #35, opened 2026-01-14
Nginx config example


Synapsis8:
Has anyone actually used the provided nginx configuration?

I'm not an nginx expert, but after messing with it about an hour I haven't got it to work.
One of the problems I ran into is that nested if statements are not allowed in nginx config files...
I managed to fix that, but then it's still throwing up random 302 & 404 errors.

PR #388, opened 2026-08-27
perf(cache): single-backend QueryCache + O(1) generation invalidation + instrumentation (#387)

fabiodalez-dev:
First measured, low-risk block of the caching overhaul tracked in #387 (steps 1-3). Redis and edge/full-page cache are deliberately deferred to later PRs pending the numbers.

## What changed
- **QueryCache — single backend** (APCu when available+enabled, else file): no more mandatory dual-write to disk; the file `.lock` is used only on the file backend, an `apcu_add()` sentinel guards stampede on the APCu hot path.
- **O(1) namespace invalidation via generation counters**: `ContentCache::booksChanged()/homeContentChanged()` bump a counter instead of scanning `APCUIterator`+`glob`. The storage key embeds the namespace generation; counters are monotonic (init to `time()`) so a lost counter can't resurrect stale entries. Unknown prefixes keep the legacy scan.
- **Instrumentation**: `QueryCache::stats()` → `{backend, gets, hits, misses, hit_ratio}` from per-request counters (zero alloc/log per call). No `index.php` wiring yet — a later PR surfaces it.
- **OPcache guidance only** (`php.ini.recommended`, no runtime change): removed `opcache.fast_shutdown` (no-op since 7.2), documented why `validate_timestamps=0` needs deploy-side invalidation, JIT left off (DB/IO-bound), 128/10000 kept with a "raise only on `opcache_get_status()` saturation" note.

## Safety / existing installs
- No migration, no new config, no new PHP extension. Non-namespaced keys hash identically to before (cache survives upgrade); namespaced keys miss once, stale files reclaimed by `gc()`/APCu TTL.
- Security invariants preserved: file cache keeps `unserialize(allowed_classes=false)` + expiry; `pinakes_` isolation kept; F008 stampede behavior preserved.
- **Known trade-off** (single backend, as designed): a server with APCu in FPM but not CLI no longer partially propagates invalidations across that boundary — already the case with the old APCu-first read, bounded by short data TTLs; availability stays live outside the cache.

## Tests
- New `tests/querycache-backend-and-generation.unit.php` (30 checks): round-trip, TTL expiry, remember-once, single-backend evidence, generation invalidation (unrelated namespace survives), object-payload rejection, stats deltas. **Fails by design** on the pre-refactor code.
- Full unit suite **137/137**, PHPStan level 5 clean, soft-delete guard clean.

Closes part of #387.

## Summary by CodeRabbit

- **Miglioramenti**
  - Ottimizzata la gestione della cache con invalidazione più rapida e minore impatto sulle prestazioni.
  - Migliorata la coerenza dei contenuti durante gli aggiornamenti e la rigenerazione della cache.
  - Rafforzata la protezione contro richieste duplicate durante la ricostruzione dei dati.
  - Aggiunte statistiche sugli accessi riusciti e mancati alla cache.
  - Migliorata la selezione automatica del sistema di archiviazione della cache, con fallback affidabile.

- **Configurazione**
  - Aggiornata la documentazione delle impostazioni OPcache, inclusi limiti, validazione dei timestamp e JIT.

### DIFF
--- a/app/Support/ContentCache.php
+++ b/app/Support/ContentCache.php
@@ -15,15 +15,16 @@ final class ContentCache
     private static bool $booksInvalidationDeferred = false;
 
     /**
-     * A book (or its availability) changed: clear catalog counts/facets,
+     * A book (or its availability) changed: invalidate catalog counts/facets,
      * every home entry (home_page_data_v1 and home_api_count_*), and the
-     * cached genre tree.
+     * cached genre tree. O(1) generation bumps — previously cached entries in
+     * these namespaces become unreachable immediately, no scan/delete storm.
      */
     public static function booksChanged(): void
     {
-        QueryCache::clearByPrefix('catalog_');
-        QueryCache::clearByPrefix('home_');
-        QueryCache::clearByPrefix('genre_tree_');
+        QueryCache::bumpGeneration('catalog_');
+        QueryCache::bumpGeneration('home_');
+        QueryCache::bumpGeneration('genre_tree_');
     }
 
     /**
@@ -31,7 +32,7 @@ final class ContentCache
      * in an outer transaction cannot safely invalidate immediately: another
      * request could rebuild the cache from pre-commit rows. Shutdown runs after
      * the controller has committed (or rolled back), and duplicate calls from a
-     * batch collapse into a single prefix sweep.
+     * batch collapse into a single invalidation.
      */
     public static function deferBooksChanged(): void
     {
@@ -51,6 +52,6 @@ final class ContentCache
      */
     public static function homeContentChanged(): void
     {
-        QueryCache::clearByPrefix('home_');
+        QueryCache::bumpGeneration('home_');
     }
 }
--- a/app/Support/QueryCache.php
+++ b/app/Support/QueryCache.php
@@ -581,6 +581,12 @@ class QueryCache
         if (self::hasApcu()) {
             $iterator = new \APCUIterator('/^pinakes_/');
             foreach ($iterator as $item) {
+                // Never remove a live mutex: deleting it while its owner is
+                // computing would allow a second worker to acquire a new
+                // lock for the same key. The sentinel expires on its own.
+                if (str_starts_with($item['key'], 'pinakes_lock_')) {
+                    continue;
+                }
                 if (!apcu_delete($item['key'])) {
                     $successApcu = false;
                 }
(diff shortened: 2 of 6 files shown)


fabiodalez-dev:
Review completata e finding CodeRabbit chiusi.

- Il finding fuori diff sulla pulizia delle fixture di generazioni precedenti è corretto in `22268f93`, con cleanup mirato sia APCu sia file senza flush globale.
- Verificati anche i tre finding inline già corretti (double-check lock APCu, generazioni memoizzate/coerenti, JIT esplicitamente disabilitato).
- Test: cache regressions 55/55 su APCu e file; QueryCache APCu 10/10 e file/generazioni 36/36; PHP lint e diff-check puliti.

fabiodalez-dev:
## 🔍 adamsreview --full — PR #388 (QueryCache single-backend + generation invalidation)

Ran 5 lenses (diff-local, structural/opus, project-conventions, comments, security) over `main..HEAD`, then adversarially validated every candidate against the code. **No blockers.** The concurrency/lock/generation logic holds up; CodeRabbit already passed. Findings below are hardening + documented limitations.

### 🟡 F1 — `mutateGenerationFile` truncates before write (low-severity robustness edge)
`mutateGenerationFile()` does `ftruncate($handle, 0)` **before** `fwrite`. A truncate-success/write-fail (rare I/O error mid-op) leaves the counter file empty → `currentGeneration()` re-inits to `time()`. Because bumps use `max($current+1, time())`, a generation that had climbed above wall-clock would reset *below* its previous value, making TTL-live entries stored under an earlier generation reachable again (a stale serve).
**Mitigated** by `ContentCache::deferBooksChanged()` batching (one bump per request, so the generation stays within seconds of `time()` — the stale window is tiny) and by the failure requiring a rare truncate-ok/write-fail. **Not a blocker.** Suggested hardening: write to a temp file + atomic `rename()`, so a failed write never leaves the counter empty.

### 🟡 F2 — `$generationCache` memo is never re-read (latent, not triggered today)
`currentGeneration()` returns the per-process memo unconditionally once set. Correct for short-lived FPM requests, but a long-running PHP process (daemon/queue worker) would serve its resolved generation for its whole lifetime, ignoring other-process invalidations. The CLI here is short-lived so this isn't triggered now — worth a docblock note for any future worker.

### 🟢 F3 — single-value keys stay CLI→FPM incoherent (pre-existing tradeoff, bounded)
Non-namespaced keys (settings, theme, i18n) invalidated from a CLI process (no APCu) delete only the file copy and can't evict FPM's APCu copy → stale until TTL. This is the known single-backend tradeoff; the generation mechanism only covers the 5 content namespaces. Bounded by the short data TTLs. The `delete()` comment slightly overstates "web/FPM → CLI/file coherence".

### Considered and dismissed (transparency)
- **stats() counts a cached null as a miss** — coherent by design: the whole cache treats `null` as a miss (`remember` re-computes on null), so this is consistent, not a bug.
- **test #35 exact-delta assertion vs `max(current+1,time())`** — in a burst, bumps vastly outpace elapsed seconds so `current+1 > time()` holds and the delta is exact; won't trigger in practice.
- **`apcu_cas` 64-bit token** — fine on standard 64-bit PHP; a `===` fallback covers exotic builds.
- **gc() unlink-while-locked race** — worst case is a harmless cache recompute.
- **symlink-follow / missing chmod on gen + GC marker files** — extends the pre-existing `setToFile` pattern; the marker/lock files carry no data, and `storage/cache` is per-user.

**Verdict: mergeable.** F1 is a nice-to-have hardening; F2/F3 are documented limitations, not defects.

Issue #422, opened 2026-09-09
Automatic backups are never rotated, and eventually break the next update


fabiodalez-dev:
## Symptom

Updating a long-lived installation fails during installation with a copy error naming an arbitrary file:

```
Errore nella copia del file: illuminate/support/Str.php
```

The file is a red herring: it is simply where the copy loop happened to stop. On an earlier occurrence of the same failure the name was `public/installer/assets`.

## What actually happens

From the updater log of the affected instance:

```
18:07:25  Download completato            {"size_mb": 29.68}
18:07:26  Integrità pacchetto verificata (sha256)
18:07:26  ZIP valido                     {"num_files": 4397}
18:07:30  Estrazione completata
18:07:30  Verifica struttura             {"missing": []}
18:07:31  Backup file applicazione per rollback
18:07:35  ERROR  Errore nella copia del file: illuminate/support/Str.php
```

The package is downloaded, digest-verified, extracted and structurally validated. The failure comes four seconds after the **pre-rollback backup** starts: the account runs out of disk while the updater copies the application aside, and `copy()` returns false on whichever file it had reached.

## Root cause

`BackupManager::createBackup()` writes a backup before every update and **nothing ever removes the old ones**. There is no retention, no rotation, no cap — the only `unlink()` calls in the class are for temp files and for an explicit manual delete.

On the affected instance:

- **73 backup archives**, ~4 MB each, **≈300 MB**, accumulated since June
- the installation itself is 480 MB, so the backups were the majority of its growth
- the hosting account hit its disk quota, and the next update failed while making the backup that is supposed to make updates safe

Any installation that updates regularly will reach the same state; it is only a matter of how much headroom the hosting has.

## Why the upgrade gate did not catch it

`ci-upgrade-smoke.yml` installs the previous release and upgrades it for real, which is the right test — but it runs on a fresh runner with an empty disk. A directory that fills over months of production use does not exist there. This is a genuine blind spot of that gate rather than something it missed.

## Fix

Rotation implemented in #421: the ten most recent automatic backups are kept, configurable through `backup.retention_count` (`0` disables it for setups that archive elsewhere). Only the generated `backup_*.zip` files are candidates, so an archive an administrator placed in the directory by hand is never deleted, and the rotation is best-effort — it can never turn a successful backup into a failed one.

## Worth considering separately

[...]

PR #423, opened 2026-09-09
feat(updater): space preflight and a copy error that says why

fabiodalez-dev:
The two follow-ups recorded in #422, alongside the backup rotation that already shipped in 0.7.83.

## Why

An update on a full account failed mid-copy with:

```
Errore nella copia del file: illuminate/support/Str.php
```

The package was intact — the updater had already verified its digest, its ZIP and its structure. The copy simply stopped at whatever file it had reached when the disk ran out, during the rollback backup. The name in the message was arbitrary, and it cost a healthy release: it was pulled on the assumption that the package was corrupt, which burned its tag.

## Space preflight

Runs right after the existing writability preflight and **before** the rollback copy that actually fills the disk, so an update that cannot fit is refused before anything is touched.

The requirement is measured from the same directories that copy duplicates: backup, restore and the estimate now read one shared constant, so a directory added to the backup cannot leave the estimate behind. A 30% margin covers new files landing beside the old ones during the copy.

Two checks, because on shared hosting neither is sufficient alone:

- `disk_free_space()` sees the **filesystem** — on the affected cPanel account it reported 56 GB free while the account's own quota was exhausted;
- a **real write** proves what the account can do right now. Capped at 16 MB: the point is to prove the account can write at all, not to reserve the full amount and double the cost of every update.

## A copy error that names its cause

`Errore nella copia del file: <path> — <reason>`, where the reason is: space or quota exhausted, destination file exists and is not writable, destination directory missing, destination directory not writable, or whatever PHP last reported. Never empty — an empty reason is the original defect.

## Tests

`tests/update-space-preflight.unit.php`, 13 checks against the real class: the estimate follows the shared constant (and the suite asserts all three call sites read it), the probe answers correctly for a small write and refuses one larger than the free space, each failure cause is named for its own scenario, a cause is always reported, the preflight passes on a healthy checkout, and the probe file is removed afterwards.

Existing updater suites re-run green (`updater-hardening` 37, `updater-custom-locale` 19, `backup-retention` 11). PHPStan clean, five-locale parity verified with the eight new strings translated.

## Summary by CodeRabbit

- **Miglioramenti**
  - Gli aggiornamenti verificano spazio, quota e scrivibilità prima di backup, download e installazione.
  - I controlli considerano anche la dimensione degli archivi e vengono ripetuti prima dell’installazione.
  - Gli errori indicano più chiaramente problemi di spazio, permessi, directory mancanti o copia dei file.
  - Le condizioni preliminari non soddisfatte mostrano messaggi dettagliati e risposte appropriate.

- **Compatibilità**
  - Gli aggiornamenti dall’applicazione sono disabilitati nell’immagine Docker ufficiale; è indicata la procedura alternativa.

- **Localizzazione**
  - Aggiunti i nuovi messaggi in italiano, inglese, danese, tedesco e francese.


fabiodalez-dev:
### Code review

**Branch:** `fix/update-space-preflight` → `main`
**PR:** #423 (open)
**Review ID:** `rev_20260909T215459Z6a8af0`
**Sub-agent tokens:** 4,079,324 across 27 invocations

Found 23 findings across all lanes:
- Deep lane (correctness/security): 7 resolved, 1 manual, 2 uncertain
- Light lane (ux/policy/architecture): 1 informational

---

### ✓ Auto-fixable (7)

| # | Score | Impact | File | Issue | Status |
|---|-------|--------|------|-------|--------|
| F003 | 72 | correctness | `app/Support/Updater.php:2321-2331` | The free-space preflight sits at the last and cheapest disk consumer: performUpdate() already ran createBackup() (DB dump into storage/backups) and downloadUpdate() (ZIP plus extraction into storage/tmp) before installUpdate() is called, so an account whose quota is exhausted still dies inside those earlier steps with the same undiagnosed error the PR set out to fix. | ✓ fixed and verified (`c07747b`) |
| F004 | 76 | correctness | `app/Support/Updater.php:3323-3330` | Parallel copy path not updated: copyDirectory() - the loop that copies the NEW release over the live installation, run immediately after backupAppFiles() and equally exposed to a full disk - still throws the bare 'Errore nella copia del file: %s' with no cause and still calls copy() unsuppressed, so the diagnostic added to copyDirectoryRecursive() covers only half the failure surface. | ✓ fixed and verified (`c07747b`) |
| F008 | 85 | correctness | `tests/update-space-preflight.unit.php:86-94` | The unit test physically fills the machine's filesystem: canWriteBytes() has no internal cap (the 16 MB cap lives only in its caller), so asking for free-space plus 1 GB writes 1 MB chunks until ENOSPC - on a dev box that is the documented trigger for a MySQL abort, and on a large disk it runs for a long time first. | ✓ fixed and verified (`c07747b`) |
| F015 | 62 | correctness | `tests/update-space-preflight.unit.php:112-117` | The unwritable-directory case silently inverts when the test runs as root (Docker/CI): is_writable() returns true on a 0555 directory for uid 0, so describeWriteFailure() falls through to the error_get_last() branch and the assertion fails for reasons unrelated to the code under test. | ✓ fixed and verified (`c07747b`) |
| F016 | 70 | correctness | `scripts/manual-upgrade.php:469-472` | The CLI upgrade path - the fallback the Updater's own error message points operators to - keeps a fixed 200 MB threshold and its own bare 'Copia file fallita' messages, so it gains neither the size-aware estimate nor the quota write probe added here. | ✓ fixed and verified (`c07747b`) |
| F024 | 76 | correctness | `app/Support/Updater.php:2499-2530` | The preflight can only detect 'cannot write at all', never 'cannot write enough', in exactly the scenario it was built for: the only comparison against $needed lives in the disk_free_space() branch, which the function's own docblock says is unreliable under a cPanel account quota, while the fallback write probe is capped at SPACE_PROBE_BYTES (16MB) and answers merely 'can a small file be written'. An account with, say, 50MB of quota headroom against a ~118MB requirement passes both checks and the update still dies mid-copy. Raised independently by four Wave-1 validators (F003, F007, F011, F016). | ✓ fixed and verified (`c07747b`) |
| F026 | 68 | correctness | `app/Support/Updater.php:2578-2598` | describeWriteFailure() orders its checks least-specific first: it runs the 1MB canWriteBytes() probe BEFORE the cheap is_file/is_dir/is_writable stat checks, so on a nearly-full volume a missing or unwritable destination directory is reported as 'spazio o quota esauriti' and the actionable cause is masked; the probe also writes to the very disk it is diagnosing, once per failing file inside a copy loop. Separately, its final error_get_last() fallback was EMPIRICALLY reproduced returning a stale unrelated error (a suppressed .env read from earlier in the same process) as the 'cause' of a copy failure. | ✓ fixed and verified (`d362ef0`) |

**Cross-cutting group G1:** F003 + F008 + F024 + F026 + F015 — These five all rewrite the same three-function unit — canWriteBytes(), checkFreeSpaceForUpdate(), describeWriteFailure() — plus the single new unit test, and three of them prescribe mutually contradictory changes to the same lines: F008 wants the probe hard-bounded to 16MB internally, F024 wants the 16MB cap deleted so the probe proves the full ~95MB requirement, and F026 wants the probe's bool return split into a tri-state; on top of that F003 changes checkFreeSpaceForUpdate()'s signature and adds two new call sites. Applied independently they overwrite each other. The unified fix is to rewrite the probe ONCE as a single self-bounding, tri-state primitive — probeWrite(int $bytes): string returning ''/'nospace'/'unavailable' — that (a) first compares $bytes against @disk_free_space() and refuses immediately when the filesystem provably cannot hold it, which is what makes the test's free+1GB request answer in milliseconds instead of filling the volume (F008's hazard) WITHOUT capping the guard's usefulness, (b) proves a genuinely large requirement by strided block-touching allocation rather than dense 1MB streaming, so the full estimate can be proven at a fraction of the I/O (F024), and (c) distinguishes 'the probe could not be created' from 'the write ran out of room', memoised per Updater instance so the multiple gate call sites and the 20-iteration bundled-plugin loop probe once (F026). canWriteBytes(): bool stays as a thin wrapper. checkFreeSpaceForUpdate() then takes the optional extra-bytes argument, branches on the tri-state, and is called early in performUpdate() and performUpdateFromFile() before createBackup() while the existing installUpdate() call site is KEPT, never moved (F003). describeWriteFailure() captures error_get_last() as its first statement, runs the cheap stat checks, and consults the memoised probe last. Finally the unit test is rewritten in one pass — bound-and-timing assertions replacing the disk-filling assertion, gate-placement assertions, a full-requirement probe assertion, and is_writable() preconditions guarding both chmod-manufactured cases including the new 0555 storage/tmp case F026 adds (which is root-invertible in exactly the way F015 describes). SETTLED BY THE ORCHESTRATOR: F008 and F015 disagreed on the skip-line form; scripts/ci-run-unit-tests.sh:32-40 matches '^SKIP:' at column 0 and, under CI_STRICT_TESTS=1, treats a skip as a failure by design. The indented form is therefore the dishonest one (a skip that reports as a pass); prefer eliminating the skip entirely, and where unavoidable use the honest line-initial form.
**Cross-cutting group G2:** F004 + F016 — Both findings edit scripts/manual-upgrade.php's copyTree() copy failure with the same change — swap copy() for @copy() and append a cause to the bare message — and both are instances of one invariant the PR only half-applied: every write failure surfaced during an upgrade must name its cause, not just the file the loop happened to reach. Applied separately they would produce two divergent inline cause helpers inside a file that must stay standalone (no autoloader, no translation function, no app classes), or one would silently revert the other's edit to the same lines. Fix them as one change: in Updater.php route copyDirectory()'s copy, unlink and mkdir branches and copyDirectoryRecursive()'s mkdir branches through describeWriteFailure(), reusing the two-argument locale key this PR already added to all five locales and adding the one new 'cannot create directory' key to all five in the same commit; in scripts/manual-upgrade.php add ONE local, dependency-free pair of helpers (a bounded write probe and a describeWriteFailure mirror, hardcoded Italian per that file's convention) and use it for both space checks and for the copyTree, mkdir and critical-file-backup messages. Lock the parity with a source-reading assertion in the unit test, following the precedent in tests/loan-coherence-audit.unit.php — a file the autoloader never touches has no other gate.

<details><summary>Details and fix proposals</summary>

#### F003 — The free-space preflight sits at the last and cheapest disk consumer: performUpdate() already ran createBackup() (DB dump into storage/backups) and downloadUpdate() (ZIP plus extraction into storage/tmp) before installUpdate() is called, so an account whose quota is exhausted still dies inside those earlier steps with the same undiagnosed error the PR set out to fix.

**File:** `app/Support/Updater.php:2321-2331`
**Score:** 72 (moderate)

**Evidence:**
- app/Support/Updater.php:2327 — checkFreeSpaceForUpdate() has exactly ONE call site, inside installUpdate(), immediately before backupAppFiles()/copyDirectory().
- app/Support/Updater.php:4325,4333,4341 — performUpdate() runs createBackup(), downloadUpdate(), installUpdate() in that order, so the gate runs only after the first two write-heavy steps have consumed disk.
- app/Support/Updater.php:1869,1906,1939 — the manual-upload entry point performUpdateFromFile() has the same order and reaches the gate last too.
- app/Support/Updater.php:2454-2461 — estimateUpdateSpace() sums only APP_BACKUP_DIRS x 1.3. On this checkout those dirs are 91MB, i.e. an estimate of ~118MB: the gate models the rollback copy only, not the backup archive or the downloaded/extracted package.
- app/Support/Updater.php:1431 — when the account is full, downloadUpdate() fails at file_put_contents() and throws a bare 'Impossibile salvare il file di aggiornamento', surfaced as 'Download fallito'. No mention of space or quota — exactly the undiagnosable class of error the PR set out to remove.
- app/Support/Updater.php:1578,1908 — an extraction that runs out of space throws a bare 'Estrazione del pacchetto fallita'. The ZIP plus extracted tree in storage/tmp is ~130MB on a real install, as much as or more than the ~118MB the gate models, and it lands BEFORE the gate runs.
- app/Support/BackupManager.php:111,140 — the pre-update backup (Step 1) reports quota exhaustion as 'Impossibile creare il file di backup' / 'Errore nella scrittura del backup'. Third unguarded writer, FIRST in execution order.
- app/Support/Updater.php:142-146 — the only check that DOES precede everything is the constructor's filesystem-level 200MB test, blind to a cPanel account quota — the very limitation the new docblock documents.
- Commit message: 'refuse an update that will not fit'. The implementation refuses only the last third of what the update writes; the first two thirds are still discovered by failing.

**Approach:** Add an EARLY space gate at the top of both update entry points, before createBackup(), sized for the whole update, while KEEPING the existing gate in installUpdate() — the later one re-measures after the earlier steps consumed disk and is what actually protects the copy. Then wire describeWriteFailure() into the remaining unguarded write-failure sites so that whichever step runs out of room says so, reusing the already-translated cause strings so no locale file changes are needed.

**Files to modify:**
- `app/Support/Updater.php` — Change checkFreeSpaceForUpdate() to accept an optional int $extraBytes = 0 and compute $needed = estimateUpdateSpace() + $extraBytes, keeping the <= 0 fall-through and both existing message strings unchanged. Leave the existing call site as-is. (why: Lets the same already-translated gate express a larger requirement early without duplicating logic or adding new i18n keys.)
- `app/Support/Updater.php` — In performUpdate(), inside the existing try block before createBackup(), call the gate with an allowance for the package: resolve the release once (its result is memoised, so downloadUpdate() does not pay twice) and take the pinakes-*.zip asset size; pass extraBytes = assetSize * 4 (ZIP + extracted tree + margin), falling back to the estimate when the size is unavailable. (why: Step 1 and Step 2 together write as much as or more than the rollback copy the current gate models, so on a quota-exhausted account the process dies before the gate ever executes.)
- `app/Support/Updater.php` — In performUpdateFromFile(), before createBackup(), call the gate with extraBytes derived from the uploaded ZIP's filesize (extraction roughly triples it). (why: The manual-upload route is the fallback an operator reaches for after the automatic update failed; it currently has no space gate at all before backup and extraction.)
- `app/Support/Updater.php` — Append the cause to the download and extraction failures, computing the description BEFORE the cleanup that deletes the extract path and the ZIP so the probe still sees the real state. (why: Same invariant the PR fixed for copies: a write failure must state why. The separator is punctuation and the cause strings are already translated, so the locale files stay untouched.)
- `app/Support/BackupManager.php` — Append the underlying reason to the two RuntimeException messages, either via a small private mirror of describeWriteFailure() or by including error_get_last(). Reuse the Updater's existing translated cause strings. (why: createBackup() is the FIRST writer in both update flows, so on a full account it is the most likely place to fail, and it currently reports no cause.)
- `scripts/manual-upgrade.php` — Next to the quota-blind disk_free_space() checks add a real bounded write probe, and append the failure cause to the copy failure. (why: This script is the remedy the new preflight message itself tells operators to run; if it is quota-blind, the recommended escape hatch fails the same opaque way.)
- `tests/update-space-preflight.unit.php` — Add assertions that pin the PLACEMENT, not just the helper: assert that in both entry points the gate call appears before the createBackup() call, that the gate has at least 3 call sites, and that passing a huge extraBytes returns a non-null message. (why: The current test only exercises the helper in isolation, which is why a gate wired at the last consumer still passes it; the ordering is the property that actually matters.)

**Verification:**
- grep -n 'checkFreeSpaceForUpdate(' app/Support/Updater.php — expect 4 hits: the declaration plus three call sites
- grep -n 'checkFreeSpaceForUpdate\|createBackup()\|downloadUpdate(\|installUpdate(' app/Support/Updater.php — in each entry point the gate's line number must be SMALLER than that method's createBackup() line
- grep -n 'describeWriteFailure' app/Support/Updater.php — expect the declaration plus at least 5 call sites
- git diff --name-only main..HEAD -- locale/ — no additional locale churn, proving the fix reused existing strings
- php tests/update-space-preflight.unit.php — all assertions pass including the new ordering ones
- phpstan analyse --memory-limit=512M — clean at level 5 full-tree
- scripts/reinstall-test.sh Test B (real admin-UI upgrade) to prove the added early gate does not block a legitimate update
- Reproduction: run with storage/ on a small tmpfs mount and trigger an update — before the fix it dies at 'Backup fallito' or 'Download fallito'; after, it must be refused up front with nothing written

**Edge cases to preserve:**
- estimate <= 0 must still return null and let the update proceed — the gate must never block a healthy install.
- disk_free_space() returning false or 0 must keep falling through to the capped write probe, never to a refusal.
- The probe stays capped at 16MB: the early gate must not attempt to reserve the full requirement, which would double the cost of every update and could itself exhaust the quota.
- The probe file must still be removed in the finally block on every path — an early gate that leaks probe files on a full account makes the situation worse.
- The gate must sit INSIDE the existing try block, after maintenance ownership and the lock are established, so an early refusal still runs through the catch that lifts maintenance.
- The existing gate in installUpdate() must NOT be removed: it measures free space after the backup and download have consumed it, which is the only measurement valid for the copy.
- Keep the release-lookup memoisation intact so the early estimate does not add a second GitHub API round trip.
- A false return from filesize() must degrade to the fallback estimate rather than producing a negative requirement.

**Latest fix attempt (fixrun_20260910T061544Zf60354):** fixed and verified

#### F004 — Parallel copy path not updated: copyDirectory() - the loop that copies the NEW release over the live installation, run immediately after backupAppFiles() and equally exposed to a full disk - still throws the bare 'Errore nella copia del file: %s' with no cause and still calls copy() unsuppressed, so the diagnostic added to copyDirectoryRecursive() covers only half the failure surface.

**File:** `app/Support/Updater.php:3323-3330`
**Score:** 76 (strong)

**Evidence:**
- app/Support/Updater.php:3328 — copyDirectory() still uses the unsuppressed copy() and the bare message. This is the exact string the PR set out to eliminate.
- app/Support/Updater.php:2745-2754 — the same construct inside copyDirectoryRecursive() WAS changed by this PR to @copy() plus the two-argument message with describeWriteFailure(). The diff touches only this site; copyDirectory() is untouched.
- app/Support/Updater.php:2338-2342 — installUpdate() runs backupAppFiles() (the FIXED path) and then, on the very next statement, copyDirectory() (the UNFIXED path). The backup has just consumed a full duplicate of app/config/locale/public/assets/installer/vendor, so at that moment the account is at its FULLEST point of the whole update: the unfixed path is MORE exposed to quota exhaustion, not less.
- app/Support/Updater.php:2303-2318 — verifyWritableTargets() already fails the update early for every permission-class problem in the copyDirectory target tree and does NOT look at free space. So the only mid-copy failure mode still reachable inside copyDirectory() is space/quota exhaustion — precisely the cause describeWriteFailure() was written to name.
- locale/*.json:7516 — the two-argument key was added to ALL FIVE locales by this PR while the old bare key survives at line 1591 in each, so extending copyDirectory() carries no i18n debt.
- app/Support/Updater.php:2414-2442 and app/Controllers/UpdateController.php:108 — when copyDirectory() throws, the catch rolls back and returns getMessage() verbatim to the admin UI, so the operator sees exactly the misleading bare filename issue #422 is about.
- app/Support/Updater.php:3323-3327 — the same causeless message is also emitted for an unlink failure, where no copy was even attempted.
- scripts/manual-upgrade.php:371 — the standalone fallback upgrader's copyTree() has the identical construct with no cause; this is the path an operator is explicitly told to use when the in-app updater fails.
- tests/update-space-preflight.unit.php:101-120 — the test exercises describeWriteFailure() directly but never asserts that any caller consumes it; nothing in the suite would notice the asymmetry.

**Approach:** Route copyDirectory()'s copy failure through the same describeWriteFailure() diagnostic, reusing the already-translated two-argument key and suppressing the copy() warning the same way. Extend the same treatment to the directory-creation and unlink branches of that loop and to the standalone fallback upgrader's copyTree(), so no remaining write failure in an upgrade reports a filename without a cause.

**Files to modify:**
- `app/Support/Updater.php` — At line 3328 replace copy() with @copy() and change the thrown message to the two-argument form with describeWriteFailure($targetPath), mirroring lines 2745-2754 verbatim including the explanatory comment. (why: This is the copy that installs the new release, executed immediately after backupAppFiles() consumed a full duplicate of the tree — the point at which an account quota is most likely exhausted. It is also the only mid-copy failure mode left in this loop, since verifyWritableTargets() already rejects permission problems beforehand.)
- `app/Support/Updater.php` — At line 3324, the !@unlink() branch: stop reusing the copy-failure string. Emit a distinct message naming the unlink of a symlinked target, appending describeWriteFailure($targetPath). (why: No copy has been attempted at that point, so the current message misattributes the failure; and a full quota is one of the causes that can make the unlink-and-replace sequence fail.)
- `app/Support/Updater.php` — At lines 3305 and 3312 change mkdir() to @mkdir() and append the cause via a new two-argument 'Impossibile creare directory' key. Add the new key to all five locale files in the same commit. (why: A directory creation is the first write of any new subtree in the release; on an exhausted quota it fails before any copy() is reached, producing a causeless message one line above the reported site. The unsuppressed mkdir also leaks a warning into the update response.)
- `app/Support/Updater.php` — At lines 2734 and 2741 (copyDirectoryRecursive's mkdir branches) append the same cause, reusing the key added above. (why: The PR diagnosed the copy inside copyDirectoryRecursive but not the mkdir in the same loop, so the backup path retains a causeless failure for the same quota exhaustion.)
- `scripts/manual-upgrade.php` — In copyTree() change copy() to @copy() and extend the RuntimeException with an inline cause, reusing the script's dependency-free style (is_writable checks then error_get_last()). Same at the critical-file backup copy. (why: This is the fallback upgrader an operator is explicitly directed to when the in-app update fails — the very next command a quota-exhausted admin runs. The script must stay dependency-free, so the cause is inlined rather than delegated.)
- `tests/update-space-preflight.unit.php` — Add a behavioural assertion that copyDirectory() — not just describeWriteFailure() in isolation — produces a message containing a cause. (why: The current test proves the helper works but never proves any caller uses it; that is exactly how the asymmetry between the two copy engines went unnoticed.)

**Verification:**
- grep -n "Errore nella copia del file: %s'" app/Support/Updater.php — must return NOTHING (only the two-argument form may remain)
- grep -n 'describeWriteFailure' app/Support/Updater.php — call sites inside BOTH copyDirectoryRecursive and copyDirectory, plus the mkdir branches
- grep -nE 'if \(!copy\(|if \(!mkdir\(' app/Support/Updater.php — must return NOTHING
- grep -n 'Copia file fallita' scripts/manual-upgrade.php — the message must carry a cause segment
- for f in locale/*.json; do grep -c 'Impossibile creare directory: %s' $f; done — must print the same count for all five (i18n parity)
- phpstan analyse --memory-limit=512M — level 5 clean
- php tests/update-space-preflight.unit.php — passes including the new copyDirectory assertion

**Edge cases to preserve:**
- The bare key must NOT be deleted from the locale files while any other caller still uses it.
- copyDirectory's skip/preserve semantics and the isCustomLocalePath early-continue must stay untouched — verifyWritableTargets() mirrors them exactly and any drift makes the dry-run lie.
- The prefix-collision guard and the symlink-replacement rationale must survive verbatim; the unlink branch changes only its MESSAGE, never its security behaviour.
- describeWriteFailure() writes a 1MB probe on every invocation — it must remain on the failure path only, never inside the per-file success loop.
- @copy() suppression must not swallow the outcome: the if(!...) test and the throw are what stop the update; only the PHP warning is suppressed.
- scripts/manual-upgrade.php must stay dependency-free before its autoload — do not introduce __() into copyTree().
- The Windows path normalisation around the copy call must be preserved.

**Latest fix attempt (fixrun_20260910T061544Zf60354):** fixed and verified

#### F008 — The unit test physically fills the machine's filesystem: canWriteBytes() has no internal cap (the 16 MB cap lives only in its caller), so asking for free-space plus 1 GB writes 1 MB chunks until ENOSPC - on a dev box that is the documented trigger for a MySQL abort, and on a large disk it runs for a long time first.

**File:** `tests/update-space-preflight.unit.php:86-94`
**Score:** 85 (strong)

**Evidence:**
- tests/update-space-preflight.unit.php:90 — invokes the private canWriteBytes() via reflection with (int) disk_free_space($root) + 1GB, deliberately larger than every free byte on the volume.
- app/Support/Updater.php:2537-2572 — canWriteBytes() contains no cap: it loops `while ($written < $bytes)` writing real 1MB chunks (explicitly non-sparse by design) until fwrite() returns false/0. With $bytes > free space the only exit is ENOSPC.
- app/Support/Updater.php:2515 — the 16MB bound exists only in the caller: $probeBytes = (int) min($needed, self::SPACE_PROBE_BYTES). The bound is caller discipline, not a property of the probe.
- app/Support/Updater.php:2582 — the only other caller, describeWriteFailure(), also passes a bounded 1MB, so production never exceeds the cap; the destructive behaviour is reachable today only from the new test, but it is reachable and the test runs unconditionally.
- scripts/ci-run-unit-tests.sh:17,31 — the runner globs tests/*.unit.php and executes every match; there is no allow-list to opt out of.
- .github/workflows/ci-quality.yml:244-290 — CI writes a .env so the test's DB guard succeeds and the test does NOT bail early. The disk-filling write executes on every CI run, on the same host volume backing the MySQL service used by the following steps.
- scripts/ci-quality-local.sh:189-194 — the local pre-push gate runs the same script, so the same write happens on the dev Mac where free space is hundreds of GB: minutes to tens of minutes of writing before it can return false.
- app/Support/Updater.php:2566-2570 — the finally block unlinks the probe, so the fill is transient on a clean exit; it is NOT transient if the process is killed mid-write, leaving a multi-GB storage/tmp/.space_probe_* inside the checkout.
- app/Support/Updater.php:143, 4150-4155 — in-repo precedent: every other free-space decision in this class READS disk_free_space() and compares. The new probe is the only place that proves capacity by consuming it.
- tests/update-space-preflight.unit.php:87 — $huge = PHP_INT_MAX; is assigned and never used, reading as a fossil of an earlier draft that passed PHP_INT_MAX directly.
- Static reasoning on termination: for a regular file there is no short-write path other than ENOSPC/quota, and PHP CLI has max_execution_time=0, so nothing bounds the loop before the volume is full.

**Approach:** Make the probe bounded inside canWriteBytes() instead of relying on callers: above SPACE_PROBE_BYTES, answer from disk_free_space() (refusing when free space is smaller or unknown) and cap the actual write at the constant, so the function can never write more than 16MB no matter what it is asked for. Then rewrite the test's assertion so it proves the refusal is bounded and immediate rather than proving it by filling the volume.

**Files to modify:**
- `app/Support/Updater.php` — In canWriteBytes(), before opening the probe: if $bytes > self::SPACE_PROBE_BYTES, read $free = @disk_free_space($dir); return false when $free is not a positive float or when $free < $bytes; otherwise clamp $bytes = self::SPACE_PROBE_BYTES and let the bounded real write proceed. Keep the non-sparse chunk loop, the fflush check and the finally cleanup unchanged. (why: The helper's worst case is 'fill the volume'. Bounding it inside the function removes the hazard for every present and future caller, keeps both production call sites byte-for-byte identical in behaviour, and preserves the cPanel quota-detection property: when the filesystem reports plenty free but the account quota is exhausted, execution still falls through to the real write.)
- `app/Support/Updater.php` — Extend the docblock to state the new invariant: the probe never writes more than SPACE_PROBE_BYTES, and a request above that is answered from disk_free_space() (unknown free space = refusal). (why: The current comment explains why the write is non-sparse but says nothing about an upper bound, which is exactly the assumption the new test violated.)
- `tests/update-space-preflight.unit.php` — Replace the section-B block: drop the unused $huge = PHP_INT_MAX; and assert the bound instead of the fill — call canWriteBytes(free+1GB) while measuring microtime and disk_free_space before/after, asserting it returns false, completes in well under a second, and free space did not drop by more than SPACE_PROBE_BYTES. Add a structural assertion reading SPACE_PROBE_BYTES via ReflectionClassConstant. (why: Without changing the test the fix is unverified, and with the fix in place the old assertion would still be the only thing between a future refactor and a filled disk.)
- `tests/update-space-preflight.unit.php` — Change the skip line from an indented '  SKIP disk_free_space() unavailable...' to a line starting at column 0 with 'SKIP:'. (why: scripts/ci-run-unit-tests.sh matches ^SKIP:; the indented form is invisible to CI_STRICT_TESTS=1, so the branch would pass as green while asserting nothing — the fake-green pattern this repo has already been bitten by.)

**Verification:**
- grep -n 'SPACE_PROBE_BYTES' app/Support/Updater.php — expect a hit INSIDE canWriteBytes(), not only at the definition and the caller
- read app/Support/Updater.php:2537-2575 — confirm no path reaches the while loop with $bytes > self::SPACE_PROBE_BYTES
- grep -rn 'canWriteBytes' app tests — every call site bounded by the caller or provably bounded by the internal clamp
- grep -n 'PHP_INT_MAX' tests/update-space-preflight.unit.php — expect no match
- grep -n '^SKIP:' tests/update-space-preflight.unit.php — the skip branch must match the runner's strict pattern
- phpstan analyse --memory-limit=512M — zero errors full-tree
- Only after the clamp is in place: php tests/update-space-preflight.unit.php, confirm section B completes in under a second and df -h shows a delta under 16MB

**Edge cases to preserve:**
- The probe must keep writing REAL bytes in 1MB chunks — a sparse allocation consumes no quota, which is what it detects.
- cPanel case: filesystem reports gigabytes free while the account quota is exhausted — a request at or below 16MB must still perform the actual write and return false, so the clamp must not short-circuit bounded requests.
- describeWriteFailure()'s 1MB probe must still return false on a genuinely full disk so the operator gets the space message rather than a vendor file name.
- canWriteBytes() must keep returning false when storage/tmp cannot be created or fopen() fails.
- The probe file must still be removed on every path including exceptions.
- checkFreeSpaceForUpdate() must keep returning null on a healthy checkout — the clamp must not turn the preflight into a gate that blocks legitimate updates.

**Latest fix attempt (fixrun_20260910T061544Zf60354):** fixed and verified

#### F015 — The unwritable-directory case silently inverts when the test runs as root (Docker/CI): is_writable() returns true on a 0555 directory for uid 0, so describeWriteFailure() falls through to the error_get_last() branch and the assertion fails for reasons unrelated to the code under test.

**File:** `tests/update-space-preflight.unit.php:112-117`
**Score:** 62 (moderate)

**Evidence:**
- tests/update-space-preflight.unit.php:112-117 — the assigned site depends entirely on is_writable() honouring the mode bits.
- app/Support/Updater.php:2578-2598 — after canWriteBytes() succeeds, describeWriteFailure() tests is_file && !is_writable, !is_dir, !is_writable($dir), then falls through to error_get_last(). Under root every permission branch is dead, so the fall-through is always taken.
- EMPIRICAL, root (php:8.2-cli container): uid=0, is_writable(0444 file)=true, is_writable(0555 dir)=true — confirms the premise.
- EMPIRICAL, non-root (uid 501): both return false. The test passes only because it is normally run as a non-root user — environment-dependent, not universally broken.
- EMPIRICAL, simulated root fall-through: replaying the branch order with a prior suppressed @file_get_contents() failure returned 'Failed to open stream: No such file or directory' for BOTH cases, so both assertions fail.
- tests/update-space-preflight.unit.php:105-110 — the SAME invariant break one block earlier (chmod 0444 + assert 'non e scrivibile'), so a root run produces two false failures, not one.
- .github/workflows/ci-quality.yml:43,283-290 — the unit-test job is runs-on: ubuntu-latest with no container: key, so it executes as the non-root runner user. CI is GREEN today.
- No Dockerfile is committed to this repo; the image is built in a separate repo and does not run the unit suite, which caps the practical blast radius.
- IN-REPO PRECEDENT: tests/updater-preflight-writable.unit.php:68-72 already guards this exact hazard with a root check. The convention is established; the new test simply omits it.

**Approach:** Make the two chmod-manufactured assertions conditional on the OS actually having honoured the chmod, instead of assuming it. Guard each with the precondition it depends on (!is_writable(...)) and emit an indented SKIP notice otherwise, so a root run reports 'precondition unavailable' rather than a false failure, while the two root-safe cases keep running and the file still exits 0.

**Files to modify:**
- `tests/update-space-preflight.unit.php` — Wrap BOTH permission-manufactured cases in a precondition guard: only run each describeWriteFailure assertion when !is_writable() on the manufactured target; otherwise emit an INDENTED skip notice. Keep the chmod restores OUTSIDE the guards so the cleanup still works in both branches. Use the indented '  SKIP ' form already used in this file — a line-initial 'SKIP:' would be counted as a failure by scripts/ci-run-unit-tests.sh under CI_STRICT_TESTS=1. (why: is_writable() returns true for uid 0 on both a 0444 file and a 0555 directory, so under root describeWriteFailure() skips every permission branch and returns a stale error_get_last(). Both assertions then fail for reasons unrelated to the code under test and the file exits 1. The precondition guard is strictly more general than a posix_geteuid() check — it also covers filesystems that ignore mode bits and builds without ext-posix, which is not declared in composer.json.)

**Verification:**
- grep -n 'chmod(' tests/update-space-preflight.unit.php — every chmod that manufactures a failure must be followed by an is_writable()-based guard; the restores must remain unguarded
- grep -n 'is_writable\|SKIP' tests/update-space-preflight.unit.php — expect two new guards and two indented SKIP notices
- grep -nE '^SKIP:' tests/update-space-preflight.unit.php — must return nothing
- Root run in a Linux container: exit code 0 with the two SKIP notices, other sections still executing
- Non-root run: exit code 0 with NO skip notices — the guard must not silently disable the assertions where they are meaningful
- read app/Support/Updater.php:2578-2598 — describeWriteFailure() itself must be unchanged; this is a test-only fix

**Edge cases to preserve:**
- The missing-destination-directory case and the always-reports-a-cause case are root-safe and MUST keep running — a blanket whole-file root skip would delete real coverage.
- The chmod restores must still execute on the skip path, or cleanup leaves a 0555 directory in the system temp dir.
- The file must keep exiting 0 on a clean root run; a guard implemented as a line-initial 'SKIP:' would be counted as a failure under CI_STRICT_TESTS=1.
- posix_geteuid() must not be called unguarded — ext-posix is not in composer.json's require block.
- The sections with no permission-dependent assertions must be left untouched.

**Latest fix attempt (fixrun_20260910T061544Zf60354):** fixed and verified

#### F016 — The CLI upgrade path - the fallback the Updater's own error message points operators to - keeps a fixed 200 MB threshold and its own bare 'Copia file fallita' messages, so it gains neither the size-aware estimate nor the quota write probe added here.

**File:** `scripts/manual-upgrade.php:469-472`
**Score:** 70 (moderate)

**Evidence:**
- scripts/manual-upgrade.php:469-472 — pre-flight 1b uses a fixed threshold with no write probe, so it measures the FILESYSTEM, exactly the metric that lies on a quota-exhausted cPanel account.
- scripts/manual-upgrade.php:586-603 — a SECOND check IS size-aware ($requiredBytes from ZipArchive::statIndex() + 100MB). The candidate's 'gains neither the size-aware estimate' is half wrong; what is missing is the quota probe.
- scripts/manual-upgrade.php:371-373 — if (!copy(...)) throw 'Copia file fallita: src -> dst' — the exact bare shape the PR replaced in Updater.php, naming whatever entry the loop reached and saying nothing about the cause.
- app/Support/Updater.php:2316 — the updater's own preflight message routes the operator to this script ('esegui l'aggiornamento da riga di comando ... php scripts/manual-upgrade.php'), translated in all five locales. The recommended fallback is the un-hardened one.
- .rsync-filter:123-132 — manual-upgrade.php is NOT excluded, so it ships inside every release ZIP: a supported, distributed upgrade path, not a dev leftover.
- scripts/manual-upgrade.php:673-691 vs :879-888 — this path backs up ONLY .env, config.local.php and version.json before copyTree() overwrites the live tree, and the catch block does no file rollback. A mid-copy abort leaves a half-updated installation — strictly worse than the Updater case, which has restoreAppFiles().
- Measured on this checkout: app 7.1MB + locale 2.7 + public/assets 23.2 + installer 1.1 + vendor 52.1 = 86.2MB, x1.3 = ~112MB. Two hard-coded gates say 200MB (Updater.php:141-146, manual-upgrade.php:469-472) and one says 100MB (Updater.php:4150-4162) — three unrelated constants for one quantity that is now computable.
- app/Support/Updater.php:141-146 — the constructor's fixed 200MB gate is BLOCKING and runs before checkFreeSpaceForUpdate(), so it can refuse an update the accurate estimate would allow.
- tests/loan-coherence-audit.unit.php:94 — in-repo precedent that already reads scripts/manual-upgrade.php source to enforce parity with the app code.

**Approach:** Give the shipped fallback the same two guarantees the Updater just gained: a real non-sparse write probe so an exhausted quota is caught BEFORE the tree is touched, and a cause-first explanation appended to every copy/backup failure. Duplicate the logic locally — the script must stay dependency-free — and lock the parity with a source assertion in the unit test, following the precedent already used for this file.

**Files to modify:**
- `scripts/manual-upgrade.php` — Add two standalone helpers next to formatBytes(): canWriteBytes($rootPath,$bytes) — mkdir -p storage/tmp, fopen wb, write real 1MB chunks up to a 16MB cap, fflush, always fclose+unlink in finally; and describeWriteFailure($rootPath,$targetPath) returning in order: quota exhausted, target unwritable, destination directory missing, destination directory unwritable, else error_get_last(), else a generic cause. Same hardcoded Italian the file already uses. (why: Without a real write the script cannot distinguish 'filesystem has room' from 'this account may still write', which is the precise confusion issue #422 was filed about.)
- `scripts/manual-upgrade.php` — In pre-flight 1b keep the disk_free_space gate but add a 16MB probe, throwing a quota-exhausted message when it fails. (why: This is the last point before the DB dump and the tree overwrite where a refusal costs the operator nothing; after it, an abort leaves a half-updated installation.)
- `scripts/manual-upgrade.php` — In the ZIP-derived check add a probe of min($requiredBytes, 16MB). (why: The mysqldump lands between the two checks and can itself be what exhausts the quota — the comment at line 586 already acknowledges the pre-flight value goes stale.)
- `scripts/manual-upgrade.php` — Change copyTree()'s failure to @copy() and append describeWriteFailure(); same for the mkdir failure and the critical-file backup failure. copyTree() already receives $rootDst, so no signature change is needed. (why: These are the messages an operator actually reads on the fallback path; leaving them bare reproduces verbatim the misdiagnosis the PR set out to eliminate — and here there is no restoreAppFiles() to undo the damage.)
- `app/Support/Updater.php` — Replace the constructor's hard-coded 200MB gate with a cheap sanity floor or drop it in favour of the computed estimate, and change checkRequirements()'s 100MB row to report estimateUpdateSpace(). (why: Three unrelated constants now co-exist with a computed ~112MB estimate. The constructor's is BLOCKING, so it can refuse an update the accurate estimate permits, and the requirements panel shows the admin a number that does not match what the preflight enforces.)
- `tests/update-space-preflight.unit.php` — Add a section asserting on the SOURCE of scripts/manual-upgrade.php (pattern already used in tests/loan-coherence-audit.unit.php:94): it defines both helpers, its copyTree failure string carries a cause suffix, and both space checks are followed by a probe call. (why: The parity is invisible to CI today, which is why the gap survived the PR; a source assertion is the only gate that works for a file the autoloader never touches.)

**Verification:**
- grep -n 'disk_free_space' scripts/manual-upgrade.php app/Support/Updater.php — every hit must be followed within ~10 lines by a probe call
- grep -n 'Copia file fallita\|Backup file critico fallito' scripts/manual-upgrade.php — each message must end with a cause suffix
- grep -n '200 \* 1024 \* 1024\|100 \* 1024 \* 1024' app/Support/Updater.php scripts/manual-upgrade.php — no hard-coded requirement contradicting estimateUpdateSpace() should remain
- php -l scripts/manual-upgrade.php — standalone file, never linted by the autoloader
- php tests/update-space-preflight.unit.php — must pass and now cover the manual script
- phpstan analyse --memory-limit=512M — new helpers level-5 clean

**Edge cases to preserve:**
- manual-upgrade.php must remain standalone: no autoload, no __(), no App\ classes — it is the tool used when the app itself will not boot.
- Its messages are hardcoded Italian by design; do NOT route them through the locale files.
- The probe must write REAL bytes — a sparse hole consumes no quota and would silently always pass.
- The probe file must be removed on every exit path and live under storage/tmp, which is in preservePaths and never overwritten by copyTree().
- CLI mode must keep working: the helpers must not depend on session or web-only state.
- The probe cap must stay bounded — probing the full requirement would double the I/O cost of every upgrade on the hosts already short on space.
- disk_free_space() returning false must keep meaning 'unknown, proceed to the probe', never 'refuse'.
- The ZIP-derived check must keep its own 100MB margin — that covers the extraction, which the app-dir estimate does not model.

**Latest fix attempt (fixrun_20260910T061544Zf60354):** fixed and verified

#### F024 — The preflight can only detect 'cannot write at all', never 'cannot write enough', in exactly the scenario it was built for: the only comparison against $needed lives in the disk_free_space() branch, which the function's own docblock says is unreliable under a cPanel account quota, while the fallback write probe is capped at SPACE_PROBE_BYTES (16MB) and answers merely 'can a small file be written'. An account with, say, 50MB of quota headroom against a ~118MB requirement passes both checks and the update still dies mid-copy. Raised independently by four Wave-1 validators (F003, F007, F011, F016).

**File:** `app/Support/Updater.php:2499-2530`
**Score:** 76 (strong)

**Evidence:**
- app/Support/Updater.php:2515 — $probeBytes = (int) min($needed, self::SPACE_PROBE_BYTES) with SPACE_PROBE_BYTES = 16MB. Since $needed is always far larger, the probe is a constant 16MB: it answers 'can this account still write 16MB', never 'can it write what the update needs'.
- Measured on this checkout (RecursiveIterator over the same APP_BACKUP_DIRS the code uses): raw 73.0MB, estimateUpdateSpace() = 94.9MB. So the quota-aware check covers 16MB of a 94.9MB requirement — about 17% of the failure window.
- app/Support/Updater.php:2493-2495 — the function's OWN docblock states that disk_free_space() 'sees the FILESYSTEM, which on a cPanel account can report tens of gigabytes free while the account's own quota is exhausted'. That is the exact hosting class of the cited incident, so on it the first branch never fires — and lines 2506-2512 are the ONLY place $needed is ever compared to anything.
- app/Support/Updater.php:2622-2629 — backupAppFiles() copies the same APP_BACKUP_DIRS (~73MB) immediately after the preflight, into the same storage/tmp the probe writes to. Any account whose remaining quota lands in [16MB, 73MB) passes both checks and then dies inside copyDirectoryRecursive() — the identical mid-copy failure the PR exists to prevent.
- tests/update-space-preflight.unit.php:87-95 — the only negative assertion for the probe is canWriteBytes(disk_free_space + 1GB) === false, which is satisfied by the FILESYSTEM, not by a quota. No test exercises 'headroom above the cap but below the requirement', i.e. the scenario the PR was written for; the suite therefore cannot detect this gap.
- Mitigating fact, verified: the gap failure lands in backupAppFiles(), and $appBackupPath is only assigned on its successful return (line 2337), so no rollback runs and no app file is touched — the install is not left half-updated, and describeWriteFailure() does emit the correct space message at that point.

**Approach:** Make the probe prove the actual requirement instead of a constant: drop the min() cap so canWriteBytes() is asked for the full $needed, keeping the existing chunked early-exit so a short account fails on the first refused write rather than after writing everything. Keep the cost negligible by allocating one byte per filesystem block (1KB stride) instead of streaming the full payload — a hole consumes no quota, but a touched block does, so a strided write allocates the whole range at a fraction of the I/O. Then extend the same real-allocation gate to the CLI upgrade path, which today has no probe at all.

**Files to modify:**
- `app/Support/Updater.php` — In checkFreeSpaceForUpdate() replace min($needed, self::SPACE_PROBE_BYTES) with the full $needed; repurpose SPACE_PROBE_BYTES as a FLOOR for tiny estimates or delete it, and rewrite the docblock, which currently documents the cap as intentional. Adjust the failure message so it stops calling 'di prova' an amount that is now the real requirement. (why: This is the only comparison against $needed that survives on quota-capped hosting; while it is capped at 16MB the gate cannot distinguish 'no space' from 'not enough space', which is the precise failure the PR cites.)
- `app/Support/Updater.php` — In canWriteBytes() add a strided allocation mode for large sizes: seek to each 1024-byte boundary and write one byte (checking every fwrite plus the final fflush), instead of streaming 1MB chunks; keep the existing dense-write behaviour for the small sizes describeWriteFailure() passes. Preserve the finally-block unlink. (why: Without it, a full-size probe writes ~95MB through PHP on every update on shared hosting — slow enough to risk max_execution_time, which is exactly the objection the current docblock raises against removing the cap. Strided writes allocate the same number of blocks (sparse holes consume no quota, touched blocks do) at a fraction of the I/O.)
- `scripts/manual-upgrade.php` — Replace the two disk_free_space-only gates with the same estimate-plus-real-allocation check: measure the directories the script backs up, then prove that many bytes can actually be written under storage/tmp before extracting or copying. (why: The CLI path is the documented fallback for exactly the hosts where this bug bites (it is named in the Updater's own error message), performs the same backup and copy, and today has no quota-aware check whatsoever — fixing only the web path leaves the same production failure fully reachable.)
- `tests/update-space-preflight.unit.php` — Add the missing case: assert that the gate probes for estimateUpdateSpace() rather than a capped constant, and assert the probe file is removed after a FAILED probe, not only a successful one. (why: The current suite passes with the capped probe in place, so it green-lights the very gap under review; without this assertion the fix can regress silently.)

**Verification:**
- grep -n 'SPACE_PROBE_BYTES' app/Support/Updater.php — must no longer appear inside checkFreeSpaceForUpdate() as an upper bound (only, if kept, as a floor)
- grep -n 'min(\$needed' app/Support/Updater.php — must return nothing
- grep -n 'disk_free_space' scripts/manual-upgrade.php — each remaining hit must be paired with a real-allocation probe call
- Reproduce on a quota-like sandbox: create a small limited volume with ~40MB free, point rootPath at a checkout, call checkFreeSpaceForUpdate() via reflection — must return a non-null message (before the fix it returns null)
- php tests/update-space-preflight.unit.php — must still report Failed: 0, with the new probe-size assertion present
- Time the healthy-install path: the preflight must add well under a second on a normal checkout, otherwise the strided allocation was not applied

**Edge cases to preserve:**
- $needed <= 0 (unmeasurable directories) must keep returning null so the gate never blocks an install it cannot measure
- The probe file must be unlinked in the finally block on every path, including a mid-write failure
- describeWriteFailure() must keep calling canWriteBytes() with its own small explicit size: after a genuine ENOSPC it needs a cheap yes/no, not a 95MB attempt
- The probe must remain non-sparse in the dense mode: a hole consumes no quota, and quota is the thing being measured
- The disk_free_space() branch must stay — it is the only check that fires when the filesystem, not the quota, is the limit

**Latest fix attempt (fixrun_20260910T061544Zf60354):** fixed and verified

#### F026 — describeWriteFailure() orders its checks least-specific first: it runs the 1MB canWriteBytes() probe BEFORE the cheap is_file/is_dir/is_writable stat checks, so on a nearly-full volume a missing or unwritable destination directory is reported as 'spazio o quota esauriti' and the actionable cause is masked; the probe also writes to the very disk it is diagnosing, once per failing file inside a copy loop. Separately, its final error_get_last() fallback was EMPIRICALLY reproduced returning a stale unrelated error (a suppressed .env read from earlier in the same process) as the 'cause' of a copy failure.

**File:** `app/Support/Updater.php:2578-2598`
**Score:** 68 (moderate)

**Evidence:**
- app/Support/Updater.php:2582 — describeWriteFailure() runs canWriteBytes(1MB) BEFORE the is_file/is_writable/is_dir checks. Reproduced against the real class: with storage/tmp chmod 0555 and an empty disk, describeWriteFailure('<root>/nope/deep/x.txt') returned "spazio su disco o quota dell'account esauriti", and so did the read-only-target case. With storage/tmp writable (control, same paths) the same two calls returned "la directory di destinazione non esiste" and "il file di destinazione esiste e non e scrivibile". Claim 1 CONFIRMED.
- app/Support/Updater.php:2537-2547 — canWriteBytes() returns false for mkdir failure, fopen failure AND write failure alike. It is a tri-state answer collapsed into a bool, so 'cannot probe' is silently rendered as 'no space' — precisely the misdiagnosis class issue #422 exists to remove.
- app/Support/Updater.php:2516 — the SAME conflation in checkFreeSpaceForUpdate() is worse than a bad message: an unwritable or uncreatable storage/tmp makes the new preflight HARD-REFUSE the whole update with the quota message, turning a permission problem into a false space block on a legitimate update.
- Claim 2 (stale error_get_last) DISPROVEN in the production path — the two claims diverge, so the score reflects claim 1. Reproduced: after a failing @copy() (exactly how line 2745 reaches line 2752), error_get_last() at line 2595 still returns the copy's own warning, because the SUCCESS path of canWriteBytes() emits no diagnostic. The stale unrelated error appears only when describeWriteFailure() is called with no preceding failing copy — i.e. tests/update-space-preflight.unit.php:120, which asserts only that the reason is non-empty and therefore passes while printing an unrelated warning. Test artefact plus latent fragility, not a live bug.
- app/Support/Updater.php:2894-2921 — the per-plugin catch in updateBundledPlugins() logs and continues instead of re-throwing, so on a genuinely full disk the 1MB probe runs once per bundled plugin. BundledPlugins::LIST has 20 entries, so one failed update can write ~20MB onto the disk it is diagnosing. Each probe is unlinked in the finally block, so nothing is left behind.
- tests/update-space-preflight.unit.php:101-116 — section C only ever exercises describeWriteFailure() with a healthy storage/tmp, which is why the ordering defect is invisible to the suite; and lines 2739-2744 create the parent directory immediately before the copy, so the !is_dir branch is effectively unreachable from the copy loop and is asserted only by the test.

**Approach:** Split canWriteBytes() into a tri-state probe ('' = ok, 'nospace' = the write itself failed, 'unavailable' = the probe could not be created at all) and let both consumers act on the distinction; in describeWriteFailure(), capture error_get_last() as the very first statement, then run the cheap stat checks, and only then consult the probe — reporting space/quota solely on 'nospace'. Memoise the probe verdict per Updater instance so the plugin loop cannot repeat it 20 times.

**Files to modify:**
- `app/Support/Updater.php` — Add a private probeWrite(int $bytes): string holding the current body of canWriteBytes() but returning '' / 'nospace' / 'unavailable' — 'unavailable' when the mkdir or the fopen fails, 'nospace' when the fwrite/fflush loop fails. Keep canWriteBytes(int): bool as a thin wrapper so the existing test assertions and PHPStan level 5 stay green. Cache the verdict in a nullable private property so the 20-iteration loop probes once. (why: The bool is a lossy encoding of a three-valued answer; every downstream misreport traces to that collapse.)
- `app/Support/Updater.php` — Rewrite describeWriteFailure(): the first statement becomes $last = error_get_last(); (before any I/O); then is_file/!is_writable target, !is_dir($dir), !is_writable($dir); then consult probeWrite(1MB), returning the space message only when it is 'nospace'; when it is 'unavailable', fall through to the captured $last message and finally to the generic cause. (why: Restores cheapest-most-specific-first ordering, removes the probe as the default answer, and makes the PHP-error fallback immune to I/O performed by the diagnosis itself — closing the latent fragility the unit test already exposes.)
- `app/Support/Updater.php` — In checkFreeSpaceForUpdate() branch on probeWrite(): keep the existing space/quota refusal for 'nospace', and add a second, distinct refusal for 'unavailable' that names the real cause and the path (storage/tmp missing or not writable, fix permissions and retry). (why: Blocking is still correct — backupAppFiles() writes into the same directory and would fail anyway — but the operator must be told to fix permissions, not to free disk space, otherwise the preflight reproduces the exact misdiagnosis it was added to eliminate.)
- `locale/it_IT.json` — Add the new key for the storage/tmp-unavailable refusal alongside the keys this PR already added. (why: Untranslated keys render as the raw msgid; the it_IT base is the source of truth.)
- `locale/en_US.json` — Same key, English translation. (why: Project rule: every user-visible string must land in every locale in the same commit — this PR already touches all five locale files.)
- `locale/de_DE.json, locale/fr_FR.json, locale/da_DK.json` — Same key in each of the three remaining locale files. (why: i18n parity is enforced by the pre-push quality gate; a key present in one locale and missing in another fails it.)
- `tests/update-space-preflight.unit.php` — In section C add a case that chmods a temp storage/tmp to 0555 and asserts describeWriteFailure() still names the directory/permission cause rather than the space message; and replace the 'always reports a cause' assertion with one that performs a real failing @copy() first and asserts the reported message contains the copy's own error — proving the fallback reports the copy's error, not a stale one. (why: The current section C only exercises the healthy-probe path, which is why the defect shipped green; without these two cases the fix is unguarded against regression.)

**Verification:**
- grep -n 'error_get_last' app/Support/Updater.php — the call inside describeWriteFailure() must be the FIRST statement of the function, above every is_file/is_dir/probe call
- grep -n 'probeWrite\|canWriteBytes' app/Support/Updater.php — expect one probeWrite() definition and three consumers; no remaining bare canWriteBytes() check that maps false directly onto a space message
- php tests/update-space-preflight.unit.php — must exit 0 with the two new section-C cases present
- Re-run the harness: instantiate Updater without its constructor, point rootPath at a temp dir, chmod <root>/storage/tmp 0555, and assert describeWriteFailure() on a missing dir and on a 0444 target returns the directory/permission messages, not the space one
- phpstan analyse --memory-limit=512M — the string-returning probe must not introduce a mixed return
- Compare key sets across all five locale files for the new msgid — all five must contain it

**Edge cases to preserve:**
- The genuine quota-exhausted case from issue #422 must still produce the space message — probeWrite() returning 'nospace' is the only path allowed to emit it
- canWriteBytes() must keep its bool signature and current semantics for the existing test assertions
- The probe file must still be unlinked in the finally block, and storage/tmp must be left free of probe files
- checkFreeSpaceForUpdate() must still return null on a healthy checkout — the new 'unavailable' branch must not fire when storage/tmp is normal
- The disk_free_space() branch must remain first, since it is free and catches the filesystem-level case before any write
- Memoising the probe must not outlive a single update run — a per-instance property is fine, a static is not

**Latest fix attempt (fixrun_20260910T093436Zddabea):** fixed and verified

</details>

### Auto-recommendations (2)

_AI-authored fix directions for findings that aren't auto-fixable today. Run `/adamsreview:fix` to batch-apply with one confirmation, or `/adamsreview:walkthrough` to review one-by-one._

| # | Score | Disp | File | Recommendation |
|---|-------|------|------|----------------|
| F018 | 60 | informational | `app/Support/Updater.php:2454-2530` | _medium:_ Memoize estimateUpdateSpace()/directorySize() results per request (an instance cache keyed by the six scanned paths) so the recursive byt... |
| F027 | 60 | manual | `app/Support/Updater.php:141-146` | _high:_ Collapse the three disagreeing space constants (the constructor's 200MB throw, checkRequirements()'s 100MB, and the true ~113MB computed ... |

<details><summary>Full recommendations and alternatives</summary>

#### F018 — The new preflight recursively walks and byte-sums six directories (including vendor/, thousands of files) via directorySize(), and unconditionally performs a real write-and-delete probe of up to 16MB on every update attempt even when disk_free_space() already confirms ample room; on the shared hosting this project targets that adds synchronous filesystem enumeration and I/O to every update, risking timeouts.

**File:** `app/Support/Updater.php:2454-2530`
**Score:** 60 (moderate)
**Reason:** Confirmed: estimateUpdateSpace() recursively walks all six APP_BACKUP_DIRS (vendor/ alone is thousands of files) and checkFreeSpaceForUpdate() always runs the up-to-16MB write probe even when disk_free_space() already confirmed room, duplicating a tree walk backupAppFiles() repeats immediately after. The design is deliberate (disk_free_space is misleading under cPanel quota) but the added synchronous I/O on time-limited shared hosting is a real architecture trade-off worth recording.
**Auto-recommendation (medium):** Memoize estimateUpdateSpace()/directorySize() results per request (an instance cache keyed by the six scanned paths) so the recursive byte-sum over vendor/ and the other dirs runs at most once per HTTP request instead of separately in checkRequirements() (page render) and again in installUpdate(); do NOT remove or gate the write-and-delete probe on disk_free_space() being ample, since that probe is what catches cPanel account-quota exhaustion that disk_free_space() cannot see — the exact scenario this PR targets.
- _Concerns:_ The hint's justification — that estimateUpdateSpace() runs redundantly in checkRequirements() and again in installUpdate() — does not match the current code: estimateUpdateSpace()/directorySize() has exactly ONE call site (checkFreeSpaceForUpdate()), invoked once from installUpdate(). checkRequirements() uses a hardcoded 100MB literal and never calls estimateUpdateSpace(). The duplication the hint assumes only exists in a hypothetical post-F027-fix world, and the hint never states that dependency.; Even granting the future combined-fix scenario, per-request memoization at best removes one duplicate call; it does not address the finding's actual claim that a single recursive walk over six directories plus a synchronous 16MB write-and-delete probe is itself expensive enough to risk a timeout on shared hosting.; Alternative B's file-cache keyed only on composer.lock changes would go stale after changes to the other backed-up directories, silently under-reporting the space requirement.
- _Alternatives:_ **B** Cache to a short-lived file entry

#### F027 — Three unrelated hard-coded space constants now coexist with the computed estimate: the constructor's BLOCKING 200MB gate (141-146) runs before checkFreeSpaceForUpdate() and can refuse an update the accurate ~112MB estimate would allow, checkRequirements() advertises 100MB to the admin (4150-4162) while installUpdate() enforces a figure roughly ten times larger, and scripts/manual-upgrade.php has its own 200MB floor. The requirements panel can show all-green on an install the preflight will refuse.

**File:** `app/Support/Updater.php:141-146`
**Score:** 60 (moderate)
**Reason:** Confirmed in substance, with two corrections. The 200MB constructor gate is PRE-EXISTING (blame: 03a8f3e1f, 2026-01-12), so this PR adds a fourth, accurate figure without reconciling the three that already existed. The arithmetic in the claim is wrong (200MB vs ~113MB is 1.8x, not ten-fold). The 'panel green, preflight refuses' contradiction is NOT reachable through the disk_free_space branch at today's install size, because the 200MB floor dominates — but it IS reachable through the quota branch, which is precisely the incident this PR exists to fix.
**Auto-recommendation (high):** Collapse the three disagreeing space constants (the constructor's 200MB throw, checkRequirements()'s 100MB, and the true ~113MB computed estimate) into one shared source of truth used everywhere, including scripts/manual-upgrade.php; downgrade the constructor's space check from a throw to a recorded warning so it stops uncaught-500ing /admin/updates (wrap all new Updater($db) call sites in UpdateController.php in try/catch as a backstop), and make checkRequirements() report a value that can never be green when checkFreeSpaceForUpdate() will later refuse the install.
- _Concerns:_ The hint never states the one mechanism the validated fix proposal identifies as closing the actual incident: adding a canWriteBytes()-driven requirement row to checkRequirements(). 'Report a value that can never be green' describes a goal, not a mechanism — an implementer could satisfy it by merely raising the disk_free_space()-compared threshold, which still cannot detect the quota-exhausted-while-filesystem-shows-space scenario.; 'Collapse into one shared source of truth used everywhere, including scripts/manual-upgrade.php' is ambiguous about whether the script should import or duplicate the constant — the validation explicitly requires its own copy, not an import of the Updater class.; The hint omits the memoisation edge case from its own verification context, even though its plan (checkRequirements() calling estimateUpdateSpace()) is exactly what introduces the double computation that edge case warns against.; The hint does not mention guarding checkRequirements()'s disk_free_space() call with @, though the verification steps require every call site to be guarded.
- _Alternatives:_ **B** Minimal blocking-bug fix only · **C** Align manual-upgrade.php only

**Evidence:**
- app/Support/Updater.php:141-146 — constructor gate: free space < 200MB throws RuntimeException. git blame shows this is pre-existing (03a8f3e1f, 2026-01-12), NOT introduced by this PR.
- app/Support/Updater.php:2454-2462 — the new estimateUpdateSpace() measured on this checkout: app 8 + config 1 + locale 3 + public/assets 24 + installer 2 + vendor 53 = 87MB, x1.3 = ~113MB. The candidate's '~112MB' is right; its 'roughly ten times larger' is wrong (200MB vs 113MB = 1.8x). The substance — three disconnected constants coexisting with one computed requirement — holds.
- app/Support/Updater.php:4150-4162 — checkRequirements() advertises required = '100MB' and marks met at >=100MB, using an UNGUARDED disk_free_space() (no @, unlike every other call site). The panel's threshold is half the constructor's blocking threshold and unrelated to the computed 113MB.
- app/Support/Updater.php:2327 — checkFreeSpaceForUpdate() runs deep inside installUpdate(), i.e. after the constructor has already run at least three times. The strictest and least informative gate always fires first; the accurate one is unreachable whenever free space sits below 200MB.
- app/Controllers/UpdateController.php:26 — new Updater($db) in index() is NOT wrapped in try/catch. With free space in the [113MB, 200MB) band the constructor throws and the entire /admin/updates page 500s: the operator cannot even reach the requirements panel that would have told them 100MB is enough. Same unguarded construction at lines 64, 95, 168, 185, 560, 626, 684.
- scripts/manual-upgrade.php:470 — a fourth constant, its own 200MB floor with independently worded message; and :594 — a FIFTH figure, uncompressedBytes + 100MB, which is the CORRECT pattern (coarse floor + payload-derived requirement) and the in-repo precedent the Updater should follow.
- Reachability: at today's install size the 200MB floor dominates the 113MB estimate, so the disk_free_space branch cannot produce the 'panel green, preflight refuses' contradiction. It IS produced by the quota branch (2515-2522): on cPanel disk_free_space() reports the filesystem (tens of GB, panel green, constructor passes) while canWriteBytes() fails on the exhausted account quota. That is exactly the incident this PR exists to fix, and the panel the operator consults is still blind to it.

**Approach:** Collapse the space checks to a single source of truth: one shared coarse floor constant for the cheap pre-install gates, checkFreeSpaceForUpdate() as the only blocking authority at install time, and checkRequirements() reporting the COMPUTED requirement plus the quota probe so the panel can never be green when the install will refuse. Downgrade the constructor's space check from a throw to a recorded warning surfaced through checkRequirements(), keeping the throw only for genuinely fatal conditions (unwritable tmp/backups, missing ZipArchive, no HTTP transport).

**Files to modify:**
- `app/Support/Updater.php` — Add a private const MIN_FREE_SPACE_BYTES next to SPACE_PROBE_BYTES. Constructor: read that constant instead of the literal, and record the shortfall in a property rather than throwing (so it no longer 500s the updates page). checkRequirements(): guard the disk_free_space() call with @, replace the fixed 100MB with max(MIN_FREE_SPACE_BYTES, estimateUpdateSpace()) for both 'required' and 'met', and add a second requirement row driven by canWriteBytes(SPACE_PROBE_BYTES) so an exhausted account quota turns the panel red. (why: This is the class holding all three disagreeing constants plus the computed estimate. Fixing only the constructor leaves the panel advertising a number no gate enforces; fixing only the panel leaves the 200MB throw shadowing the accurate check in the 113-200MB band.)
- `app/Controllers/UpdateController.php` — Wrap new Updater($db) at line 26 and the other seven construction sites (64, 95, 168, 185, 560, 626, 684) so a preflight RuntimeException renders the requirements panel with the failure explained, instead of an uncaught 500. (why: The constructor throw is the delivery mechanism for the inconsistency. While it stays uncaught, the operator on a low-space host loses the very page that would tell them what is wrong — the opposite of this PR's goal.)
- `scripts/manual-upgrade.php` — Replace the standalone 200MB literal with the same floor value used by the Updater (a named constant near MAX_ZIP_SIZE), and align its message wording with the accurate check already present further down. (why: The script is the documented recovery route when the in-app updater fails, so it must not refuse at a different threshold than the thing it is recovering.)
- `tests/update-space-preflight.unit.php` — Add a section asserting the gates agree: no bare N * 1024 * 1024 space literal remains in Updater.php outside the constant block, and checkRequirements()'s space row reports a value >= estimateUpdateSpace(). Also drop the inert second argument in new Updater($db, $root) — PHP silently ignores extra args to a one-parameter userland constructor, so the test is not rooted where it claims to be. (why: Section D only proves the preflight passes on a healthy checkout; nothing currently pins the three thresholds together, so they will drift apart again on the next edit.)

**Verification:**
- grep -nE '(200|100) \* 1024 \* 1024' app/Support/Updater.php scripts/manual-upgrade.php — must return only the shared constant definitions, no inline literals
- grep -n 'MIN_FREE_SPACE_BYTES' app/Support/Updater.php scripts/manual-upgrade.php — must appear in the constructor gate, checkRequirements() and the manual-upgrade floor
- grep -n 'disk_free_space' app/Support/Updater.php — every call must be @-guarded, including the one in checkRequirements()
- grep -n 'new Updater(' app/Controllers/UpdateController.php — every site must sit inside a try/catch
- php tests/update-space-preflight.unit.php — must exit 0 and include the new threshold-agreement assertions
- Read checkRequirements() and confirm the space row's 'required' derives from estimateUpdateSpace(), not a literal

**Edge cases to preserve:**
- estimateUpdateSpace() returning 0 must still let the update proceed — a fresh or unmeasurable tree is not a refusal
- The quota write probe must stay capped at SPACE_PROBE_BYTES; calling checkRequirements() on the admin page must not write the full 113MB
- estimateUpdateSpace() walks vendor/ recursively — memoise it per request, since checkRequirements() is now on the page-render path and is called again by performUpdate()
- disk_free_space() reporting the filesystem rather than the cPanel account quota is the whole reason the write probe exists: never replace the probe with a disk_free_space comparison
- manual-upgrade.php is standalone — it needs its own copy of the constant value, not an import of the Updater class

</details>

### ⚠ Requires manual attention (1)

_Not auto-applied by `/adamsreview:fix` directly — these need a confirmation step. Findings with an auto-recommendation get batch-confirmed at `:fix`'s Phase 7.5 preflight (or `:walkthrough` Step 4.5); use `/adamsreview:promote <finding_id>` for a single-finding manual override._

| # | Score | Impact | File | Issue | Why manual |
|---|-------|--------|------|-------|-------------|
| F027 | 60 | correctness | `app/Support/Updater.php:141-146` | Three unrelated hard-coded space constants now coexist with the computed estimate: the constructor's BLOCKING 200MB gate (141-146) runs before checkFreeSpaceForUpdate() and can refuse an update the accurate ~112MB estimate would allow, checkRequirements() advertises 100MB to the admin (4150-4162) while installUpdate() enforces a figure roughly ten times larger, and scripts/manual-upgrade.php has its own 200MB floor. The requirements panel can show all-green on an install the preflight will refuse. | Confirmed in substance, with two corrections. The 200MB constructor gate is PRE-EXISTING (blame: 03a8f3e1f, 2026-01-12), so this PR adds a fourth, accurate figure without reconciling the three that already existed. The arithmetic in the claim is wrong (200MB vs ~113MB is 1.8x, not ten-fold). The 'panel green, preflight refuses' contradiction is NOT reachable through the disk_free_space branch at today's install size, because the 200MB floor dominates — but it IS reachable through the quota branch, which is precisely the incident this PR exists to fix. |

### ℹ Uncertain (2)

| # | Score | Impact | File | Issue |
|---|-------|--------|------|-------|
| F001 | 48 | correctness | `app/Support/Updater.php:2446-2454` | The pre-existing docblock "Backup application files for atomic rollback" was left in place while new methods (estimateUpdateSpace, directorySize, checkFreeSpaceForUpdate, canWriteBytes, describeWriteFailure) were inserted between it and its target function; the comment now sits directly atop estimateUpdateSpace() instead, and the real backupAppFiles() function is left with no docblock at all. |
| F007 | 52 | correctness | `app/Support/Updater.php:2537-2547` | Both the preflight and the copy-failure diagnosis probe a hardcoded storage/tmp while the failing write may be on a different volume: copyDirectory's own comment states that storage/ and uploads/ mounted elsewhere are a supported setup, so on those installs the probe measures the wrong filesystem or quota. |

Phase 4 couldn't confirm decisively. Re-run `/adamsreview:review` if you suspect this deserves
further investigation with fresh context.

## Light lane — ux, policy, architecture

| # | Score | Impact | File | Finding | Disposition |
|---|-------|--------|------|---------|-------------|
| F018 | 60 | architecture | `app/Support/Updater.php:2454-2530` | The new preflight recursively walks and byte-sums six directories (including vendor/, thousands of files) via directorySize(), and unconditionally performs a real write-and-delete probe of up to 16MB on every update attempt even when disk_free_space() already confirms ample room; on the shared hosting this project targets that adds synchronous filesystem enumeration and I/O to every update, risking timeouts. | informational |

## Polish — below threshold, clustered (7)

Below-gate findings (score < 45) that cluster in the same area — not worth surfacing individually, but dense enough that a human pass may catch something the pipeline filtered out.

| # | Score | File | Rough location | Concern |
|---|-------|------|----------------|---------|
| F010 | 24 | `app/Support/Updater.php` | L2499-2504 | The estimate silently disables the whole gate when it cannot measure: any environment where the app directories are unreadable (open_basedir, restricted iterator - directorySize() swallows the Throwable and returns 0) yields $needed === 0 and checkFreeSpaceForUpdate() returns null without running the write probe, skipping the check precisely on the locked-down shared hosts it targets. |
| F019 | 28 | `app/Support/Updater.php` | L2506-2513 | The fast-path disk_free_space() check requires $free > 0, so a filesystem correctly reporting exactly 0 bytes free is excluded from the descriptive 'Spazio su disco insufficiente' branch and falls through to the write-probe path instead, producing a less specific message. |
| F022 | 20 | `app/Support/Updater.php` | L2515-2522 | The write-probe failure message reads awkwardly in Italian ('Impossibile scrivere %1$s di prova' = 'Unable to write %1$s of test'), missing the noun that would make it natural; 'Impossibile scrivere un file di prova di %1$s: ...' would read correctly. |
| F025 | 40 | `app/Support/Updater.php` | L2537-2548 | random_bytes(4) at the top of canWriteBytes() sits outside the try/finally and outside any catch, so a \Random\RandomException escapes the declared ': bool' contract. Both callers assume a boolean; from describeWriteFailure() the throw would surface WHILE ALREADY BUILDING an exception message for a failed copy, replacing the actionable diagnostic with an unrelated entropy error. |
| F012 | 36 | `app/Support/Updater.php` | L2543 | Probe files are not covered by the temp sweeper: cleanupOldTempDirs() globs only pinakes_update_* and pinakes_app_backup_* with GLOB_ONLYDIR, so a request killed mid-probe leaves a .space_probe_* file of up to 16 MB in storage/tmp - on the very disk the check exists to protect - with nothing that ever removes it. |
| F005 | 38 | `app/Support/Updater.php` | L2561-2571 | The write probe can report success on a full disk: fflush() only pushes PHP's userland buffer into the OS, and the return value of fclose() - where a delayed-allocation ENOSPC actually surfaces - is discarded inside the finally block, so canWriteBytes() returns true for data that never reached the filesystem. |
| F023 | 32 | `app/Support/Updater.php` | L2578-2598 | describeWriteFailure() falls back to embedding the raw PHP error_get_last() message verbatim into an exception surfaced to the admin UI, which can leak absolute server filesystem paths and internal environment details; this is the first place in Updater.php where a raw system error string is placed directly into a user-facing message rather than only into debugLog(). |

## Fix runs

### Run `fixrun_20260910T061544Zf60354` — 2026-09-10T07:07:27Z

- Outcomes: 6 fixed and verified, 1 partial
- Commits: `c07747b`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F003 | FG-1 | ✓ fixed and verified |  |
| F004 | FG-1 | ✓ fixed and verified |  |
| F008 | FG-1 | ✓ fixed and verified |  |
| F015 | FG-1 | ✓ fixed and verified |  |
| F016 | FG-1 | ✓ fixed and verified |  |
| F024 | FG-1 | ✓ fixed and verified |  |
| F026 | FG-1 | ⚠ partial | The error_get_last() hoist is correct inside describeWriteFailure() (Updater.php:2752-2779, proven at runtime by the new section-C case). But the fix applies the same principle inconsistently at the four new call sites it added. At Updater.php:1240-1247 (downloadUpdate temp mkdir) and 2793-2800 (backupAppFiles mkdir) it deliberately hoists the describeWriteFailure() call ABOVE debugLog(), with a comment stating that debugLog() writes to disk and would replace the error error_get_last() has to report — a premise verified as true: debugLog() -> SecureLogger::log() ends in @file_put_contents(storage/logs/app.log) at SecureLogger.php:39, and an @-suppressed failure still replaces error_get_last(). At the other two new sites the fix does the opposite: Updater.php:1447-1458 calls debugLog('ERROR','Impossibile salvare file') first and only then describeWriteFailure($zipPath); Updater.php:1593-1610 calls $zip->close() and debugLog first, and only then describeWriteFailure($extractPath). On a read-only or otherwise failing storage/logs (where the log write itself errors and the 1MB probe does NOT return 'nospace', so the chain reaches the error_get_last() fallback), the reported cause is the app.log write error rather than the ZIP/extraction write error — exactly the stale-unrelated-error class F026 was raised about. Impact is narrow (the ENOSPC case is caught earlier by the probe), which is why this is partial and not a regression: all of F026's listed files were edited, the tri-state probe, the 'unavailable' refusal, the five locale keys and both new section-C cases are present and pass, the memo is a per-instance array (not static), and checkFreeSpaceForUpdate() always passes fresh=true and clears it immediately before backupAppFiles()/copyDirectory()/updateBundledPlugins(), so the non-rethrowing plugin loop genuinely probes once. |

### Run `fixrun_20260910T093436Zddabea` — 2026-09-10T09:34:36Z

- Outcomes: 1 fixed and verified
- Commits: `d362ef0`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F026 | FG-1 | ✓ fixed and verified |  |

---

fabiodalez-dev:
@coderabbitai full review

